PR #15, opened 2019-03-05
Fixed implicit fallthrough in switch case

nhatzHK:
Fixed implicit fall-through in switch..case. This doesn't fix a bug, but makes it possible to compile with `-Werror`


Jason2605:
Thanks :)

PR #105, opened 2020-02-03
Use code blocks in the example column.

LamboCreeper:
Makes it so that the example functions on the built-in functions page are wrapped in code blocks.


Jason2605:
Nice addition, thanks for this 😄

PR #131, opened 2020-02-14
Adds instructions for adding a Dictu CLI command.

LamboCreeper:
Allows you to simply type `dictu <file>` instead of `<path/to/dictu> <file>`.


Jason2605:
Thanks for the PR! I am unfortunately going to close this PR and heres my reasoning. When exporting, the exported value is only available during the lifetime of the terminal session, this means if you were to open a new terminal (or close and reopen the current), the export would no longer be there, and you'd have to do it again.

This does spark some potential changes we could make instead however. One being, as part of the "make" process, we add a symlink to the `/usr/local/bin` directory.

PR #147, opened 2020-03-02
Make building the http class optional, as it requires libcurl (for no…

agathoklisx:
This allows for an optional building for HTTP class.

To disable it run:

make DISABLE_HTTP=1

and when running the tests use:

./dictu tests/runTests.du --disable-http

This is tested with GNU-make.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
-      - name: Make dictu and run tests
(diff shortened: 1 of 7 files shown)


Jason2605:
For a bit of clarity - I have updated this PR to also build Dictu in an environment that doesn't have cURL installed using the `DISABLE_HTTP` flag you have provided as part of the CI/CD process! I've also added build steps to the README file incase someone wishes to build without cURL on their machine.

Thanks again for the PR!

PR #154, opened 2020-03-23
Is defined

### LINKED ISSUE #153: Add an isDefined() builtin
# Summary
The HTTP library is currently an optional extra when building the VM - Due to this (and potentially others in the future) there needs to programatically be a way to check if a function is available at runtime
## Example
```
if (isDefined("HTTP")) {
    HTTP.get(...);
} else {
    print("HTTP disabled!");
}
```
}

agathoklisx:
# isDefined()
Resolves #153 
## Summary
This PR adds a new `isDefined()` builtin function. This function checks to see if a given value has been defined.

**Note:** that this performs a lookup only to the global table for object references, and which
is the common usage.


Jason2605:
Thanks very much for this PR! For visibility, I've merged develop into your PR and added tests and documentation for the `isDefined` builtin.

PR #169, opened 2020-04-02
Add Path Native class.

agathoklisx:
Introduce Path Native Class.


Jason2605:
This is great, thank you very much for the contribution here!

PR #175, opened 2020-04-05
Add System.platform constant, that identifies the underlying system p…

agathoklisx:
Add System.platform. For systems with uname(2) this is uname().sysname with the first
char in lower case.


Jason2605:
Thanks very much for your contribution!!

PR #176, opened 2020-04-06
Len method

agathoklisx:
Add len() method to String, Set, Dict and List datatypes.


Jason2605:
For starters, thanks very much for the contribution! I've pushed some changes to your PR, for visibility i'll write down my changes here:
- Reshuffled the len* natives to match the order they are added to the hashtable
- Removed the type check for string in lenString. Reasoning being, at the point it hits this native, we know the type must be a string
- Moved back the concatenation header on the strings documentation page
- Added a couple more tests for the `.len()` method on all types

PR #183, opened 2020-04-08
Add toString() method for ListType datatype.

agathoklisx:
This is to add the toString() method to the standard datatypes. This first commit is about
the ListType datatype.


Jason2605:
Many thanks for your work on this!

PR #184, opened 2020-04-08
Native methods

Jason2605:
# Native methods
## Summary
This PR adds the scaffolding for more native values on booleans, numbers and nil. This will compliment #183 which is currently open, in the transition from a native builtin to methods on values.


Jason2605:
Branched from the wrong place, closing. Will open a new PR. Oops.

PR #192, opened 2020-04-17
Add __file__ variable that holds the path name of the compilation unit.

agathoklisx:
Add  __file__.


Jason2605:
Thanks for this!

PR #193, opened 2020-04-18
Add 'not' keyword.

agathoklisx:
Add  'not' keyword.


Jason2605:
Great spot with this one, and thanks for the implementation!

PR #194, opened 2020-04-18
Do not raise an error in setCWD() if the operation is not successful.

agathoklisx:
Mimic chdir() and libc conventions and return 0 uppon success or -1 otherwise.
Define those values as OK and NOTOK respectively in vm.h.


Jason2605:
Thanks for this!

PR #197, opened 2020-04-28
Add System.remove() method, that deletes a file from the filesystem.

agathoklisx:
Returns 0 upon success and -1 otherwise.


Jason2605:
Thanks for this!

PR #198, opened 2020-04-28
Add System.mkdir() and System.rmdir().

agathoklisx:
Both return 0 upon success and -1 otherwise.

System.mkdir() can take an optional number argument that specifies the mode.

Add also S_I* constants to specify the permissions bits.


Jason2605:
Thanks for this!

PR #201, opened 2020-04-29
Add System.[getpid(),getppid(),getuid(),geteuid(),getgid(),getegid()]

agathoklisx:
These functions should always return succesfully.


Jason2605:
Thanks for this, once again!

PR #206, opened 2020-05-03
Handle errors from native classes.

agathoklisx:
Introduce a specific to every class "errno" variable.
Introduce a specific to every class a strerror() method.
Introduce a C class to expose libc errors.


Jason2605:
Massive thank you for the work you've put in for this PR!

PR #219, opened 2020-06-09
Suppress the compiler warnings for unused parameters.

agathoklisx:
# Summary
This PR removes the `-Wno-unused-parameter` compiler flag and adds a new `UNUSED` macro. This will aid in development as it will force the developer to explicitly set a parameter to UNUSED when it is not being used.


Jason2605:
Nice addition, thanks for this!

PR #227, opened 2020-06-21
Fix strftime() native method.

agathoklisx:
Return an empty string when a zero length string is given as an argument. Use a maximum
iterator, to catch up some cases when strftime() returns an empty string in some locales.


Jason2605:
Thanks for this!

PR #268, opened 2020-10-01
Implement list sorting for numbers.

### LINKED ISSUE #264: Ability to sort lists
# Summary
Currently there is no method to sort lists. For this task we need to add `list.sort()` which would sort a list **in place**. There are methods in place for copying lists if the requirement is to not sort in place. Initially this can be a very simple implementation and only work on certain datatypes (initially numbers) and we can expand later if required. If a list contains something that is not a number, an error will be raised.

## Example
```
var x = [1, 5, 2, 3];
x.sort();
print(x); // [1, 2, 3, 5] 
```

NightShade256:
Implement list sorting for numbers as in #264 

Resolves #264 

Changes
-----------

1. Added list sorting for numbers, with quick sort (Hoare partition method) being the sorting algorithm.
2. Added relevant tests.

Details
--------

The sorting algorithm is Quick Sort (Hoare partition scheme). The average complexity is O( n log (n) ) and the worst case complexity is O( n^ 2 ) (albeit in rare cases).

Further Improvement
------------------------

Quick sort is a bit overkill for small lists, we could use some simpler algorithm instead for list sizes less than 20 maybe.

### DIFF
--- a/c/datatypes/lists.c
+++ b/c/datatypes/lists.c
@@ -1,3 +1,5 @@
+#include <math.h>
+
 #include "lists.h"
 
 static Value toStringList(VM *vm, int argCount, Value *args) {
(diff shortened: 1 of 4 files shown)


Jason2605:
Thanks for this! I should be able to give this a test later on today and get it merged in. Before merging however we will need to add some documentation about the new method [here](https://github.com/dictu-lang/Dictu/blob/develop/docs/docs/collections.md).

NightShade256:
@Jason2605 Included a small example in the docs for this method.

Jason2605 (approved):
Perfect, thank you!

Issue #271, opened 2020-10-01
Underscore separators for numbers 


Jason2605:
# Summary
It would be a nice addition if we could allow underscores to be parsed when parsing a number, as it allows for a much more readable representation of larger numbers
### Example
```js
var myNum = 1_000_000;
```

PR #272, opened 2020-10-02
Add support for `_` as a delimiter in numbers.

NightShade256:
Implements the feature as in #271 .

Added relevant tests, and checked for leaks and such.
Note: The underscore can only occur in the non fractional part of a number, this was an intentional choice. If you want me to modify it so that it can appear even in the fractional part I will happily implement that.

### DIFF
--- a/c/compiler.c
+++ b/c/compiler.c
@@ -671,7 +671,8 @@ static void number(Compiler *compiler, bool canAssign) {
     UNUSED(canAssign);
 
     // We allocate the whole range for the worst case.
-    char* buffer = (char *)calloc(compiler->parser->previous.length, sizeof(char));
+    // Also account for the null-byte.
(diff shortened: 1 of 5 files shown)


Jason2605:
Appreciate the pull request! I'm not sure there would be many cases where the separator would be used after the decimal point however i personally think we should be consistent in allowing it on both sides not to have changing behaviour.

Something to note as well (not sure it's an "issue")
```
var x = 1_____________0;
```
Would be valid, should we consider a separator to only be a single underscore? Not sure if the extra complexity to track this would be worth the effort.

NightShade256:
For the underscore to be allowed in fractional part, consistency would indeed be nice.

For the spammy underscore issue, I think we should let it pass since other languages like Rust also allow similar behaviour.

NightShade256:
Simple fix, all done.

Jason2605:
Perfect, thank you once again!

Issue #273, opened 2020-10-03
Implement Random


Jason2605:
# Random
## Summary
Currently within Dictu there is no way to generate a random number. Along with the ability to generate random numbers it would be nice if we could both generate them between a range, and have the ability to select a random element of a list.

### Examples
```
import Random;

Random.random(); // Number between 0-1
Random.range(1, 10); // Number between start-end
Random.select([1, 2, 3]); // Randomly selected element from a list
```

SeanMcLoughlin:
Hey, mind if I take this one?

damn-dvlpr:
As per my convo with @Jason2605 , would want to work on this.

Jason2605:
Closed as part of #274

PR #274, opened 2020-10-04
Implement Random

SeanMcLoughlin:
Feature for #273. 

Still need to add `Random.range()` and `Random.select()`, but that will be done fairly quickly. 

Dictu tests are also added.

There is also a FIXME in here where I'm unsure how to surpress `-Werror` for not using a function parameter. I don't see any examples in the already existing code so I'm unsure how to cleanly get around this.

### DIFF
--- a/c/optionals/optionals.c
+++ b/c/optionals/optionals.c
@@ -7,6 +7,7 @@ BuiltinModules modules[] = {
         {"Path", &createPathClass},
         {"Datetime", &createDatetimeClass},
         {"Socket", &createSocketClass},
+        {"Random", &createRandomClass},
 #ifndef DISABLE_HTTP
         {"HTTP", &createHTTPClass},
 #endif
--- a/c/optionals/optionals.h
+++ b/c/optionals/optionals.h
@@ -11,6 +11,7 @@
 #include "c.h"
 #include "datetime.h"
 #include "socket.h"
+#include "random.h"
 
 #define GET_SELF_CLASS \
   AS_MODULE(args[-1])
--- a/c/optionals/random.c
+++ b/c/optionals/random.c
@@ -1,7 +1,7 @@
 #include "random.h"
 
 static Value randomRandom(VM *vm, int argCount, Value *args) {
-    args = args;  // FIXME: Can't get -Werror to surpress not using args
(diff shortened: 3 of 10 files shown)


TheGreatCookieMachine:
Suppression of the "Unused parameter" error is done via the `UNUSED` macro. You can see an example here: https://github.com/dictu-lang/Dictu/blob/511f268309d2e770bcefa8dade7c32efbd9f3833/c/optionals/system.c#L4
I can't test it, but simply replacing `args = args;` with `UNUSED(args);` should work.

Jason2605:
Appreciate you taking the time to do this work @SeanMcLoughlin! A couple of things to note / comment. The implementation you're using here (libc's rand) comparatively speaking is actually quite a slow algorithm, it may be worth (probably in the future) looking at other viable algorithms, such as (Mersenne Twister)[https://en.wikipedia.org/wiki/Mersenne_Twister], or one of the [XOrShift](https://en.wikipedia.org/wiki/Xorshift) family variants.

With all that being said, i'm happy to implement the `libc` variant for now, and we can swap out the internal algorithm at a later date, as long as we can ensure that the Dictu API stays the same 😄.

Some notes about the PR:
- I noticed i don't see a call to [srand](https://linux.die.net/man/3/srand) to seed the PRNG, i think a good place for this would be [here](https://github.com/dictu-lang/Dictu/pull/274/files#diff-86e991631a8ad43de27f6ddd573f63faR56-R72)
- `Random.random()` returns either 0 or 1, should this return a float between 0 or 1, or only 0 or 1. A design decision i'm open to
- We need to add some documentation for the new module and methods before this gets merged in

Outside of these points though, I'm happy you seemed to pickup the internals of the interpreter pretty quickly and was able to add this new builtin module! Thank you!

SeanMcLoughlin:
A note about the above changes: The PRNG is seeded with `time(NULL)`. This doesn't generate a proper distribution but, unless `Random` is needed for things like cryptography right now, this should hopefully suffice.

Jason2605:
That's fine, we shouldn't be using this for cryptography, i've added a note about that in the documentation. Thanks for updating the PR!

Issue #257, opened 2020-09-17
Add method to get directory contents


Jason2605:
# Directory contents
Currently within Dictu there is no way to get the contents of a directory. This is fundamental logic which will be required. I propose we add this as a new method, `listdir` as part of the `Path` library.

Issue #90, opened 2020-01-29
Build method for windows machines


Jason2605:
Currently there’s no way to build Dictu on windows.

MrFrydae:
When the user executes 'make' on a Windows system, it either installs a dual boot and compiles on linux, or installs the windows subsystem for linux and compiles using that. I am not saying that this will be a very time efficient method, but it will be very effective.

NightShade256:
Building on Windows is very much doable.

We could use [CMake](https://cmake.org/) to be OS agnostic.
I could make a `CMakeLists.txt` but it would nullify the existing testing workflow on Github Actions, so we would have to account for that.

NightShade256:
If you want me to go ahead, I will happily do so.

Jason2605:
Hi @NightShade256, I'd be very happy for you to attempt to get a build of Dictu working on Windows, we can always review the testing method on Actions - that's no problem. Obviously with this when building on windows there may be some unknowns when attempting to build so that could be quite tricky to solve at times.

Will assign this to you!

NightShade256:
@Jason2605 I think adding cURL as a git submodule will massively simplify the build experience for both the developers and the end users. If you allow it I can do it that way.

Also should I make a separate PR or since I already have one opened #268 push this as a commit to that PR?

Jason2605:
@NightShade256 Definitely something to look into if it will ease the building process.

Separate PR per item please!

NightShade256:
@Jason2605 I have hit a snag.

In `optionals/system.h` you include `sys/utsname.h` , which AFAIK is a POSIX header and hence unavailable on Windows.
(Even on MinGW-w64 for that matter)

This issue must be resolved before further progress.

Jason2605:
Yeah i had a feeling this would be the case. I'm still curious to see where you go up to with your CMake conversion however, and we can slowly work through the Dictu source to try and make it more "OS independent". Appreciate you looking into this by the way!

NightShade256:
@Jason2605 

I have actually made the `CMakeLists.txt` file, it compiles successfully on Linux currently (on my machine) for the debug and release profiles.
But since due to above issues, the scope of the port to Windows has widened from simply rewriting the build files to rewriting chunks of the interpreter itself. Plus there is the cURL dependency it would either require users to use `vcpkg` or we would have to bundle the headers and static library through this repository.

I think this issue should currently be dropped till those other problems are solved.

Jason2605:
[...]

PR #276, opened 2020-10-04
Add Path.listdir() method

SeanMcLoughlin:
For #257

### DIFF
--- a/c/optionals/path.c
+++ b/c/optionals/path.c
@@ -40,70 +40,6 @@ static Value isAbsoluteNative(VM *vm, int argCount, Value *args) {
     return (IS_DIR_SEPARATOR(path[0]) ? TRUE_VAL : FALSE_VAL);
 }
 
-static Value isdirNative(VM *vm, int argCount, Value *args) {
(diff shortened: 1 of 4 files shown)


Jason2605:
Thank you once again for another PR! A couple of points about this:
- This wouldn't work on windows systems, although we can't yet build Dictu on windows machines, there is work underway at potentially converting this (#90). It may be worth introducing some macros to cater for windows systems. Relevant library as well (https://github.com/cxong/tinydir)
- I think it would be neat if the method allowed no argument to be passed and it uses the current directory as a default

SeanMcLoughlin:
Not sure how to test this feature of defaulting to `cwd` since there's no change directory function in the `Path` library, but it's implemented.

Jason2605:
Thank you for the update @SeanMcLoughlin! I'll approve this so you get your hacktoberfest contribution (if you are taking part) however i will hold off on merging for now while #278 is in progress as this task will need updating to be windows compatible.

Jason2605 (approved):

PR #278, opened 2020-10-04
Fix compilation on Windows

TheGreatCookieMachine:
This pull request should allow Dictu to be built and run on Windows devices. This pull request builds and passes all tests on my Windows device using tdm64 gcc 9.2.0, clang 9.0.0 using MSVC's libc, and Visual Studio 16.5.4. If anyone else with a Windows machine could test the Pull Request, that would be wonderful.

## Changes to Dictu language

One change was made to the overall Dictu language and not just the Windows versions. Dictu's `file.seek` previously allowed the user to seek to an arbitrary location on a file which was opened in text mode. To do this, it used C's `fseek` function. C's `fseek` does not allow arbitrary seeks on text files:

> For text streams, the only valid values of offset are ​0​ (applicable to any origin) and a value returned by an earlier call to ftell (only applicable to SEEK_SET). - https://en.cppreference.com/w/c/io/fseek

From what I can tell, POSIX systems treat binary and text files the same, as such this issue did not pop up. Windows, on the other hand, performs special actions on text files (e.g. turning CRLF line endings into LF line endings).
I opted to restrict `file.seek` on text files on all systems to an offset of 0. If you would like me to revert `file.seek` on POSIX, let me know.

This does not affect files opened as binary.

## Removed Windows functionality

Certain features have been temporarily or permanently removed from the Windows version of Dictu.

* The advanced REPL has been removed in favor of a far simpler one. The REPL requires linenoise which in turn requires termios.h, a POSIX header. It would be possible to port linenoise to Windows headers, but that would be a big project in of itself.

* `Datetime.strptime` has been removed. `Datetime.strptime` uses C's `strptime`, which is a POSIX function. It would be possible to create a Windows version of `strptime`, but I opted to remove `Datetime.strptime` for now.

* `getgid`, `getegid`, `getuid`, `geteuid`, `getppid`, and `getpid` have all been removed. Their associated POSIX functions are not on Windows and have no alternatives on Windows. From what I can tell, Windows doesn't have process groups in the same way POSIX does.

## Internal Changes

One of the goals of how the code was set up was to affect the currently existing code and logic as little as possible. Instead of completely changing functions or adding mutual wrapper functions, I added new defines and functions which serve as wrappers around Windows functions into the POSIX API. As a result, when necessary, I added Windows specific code to the beginning of a file. Sometimes this wasn't possible, such as with VLAs.

One change that was made was replacing `_r` functions (such as `strerror_r`) with `_s` functions. The point of `_r` functions is to be thread safe. Unfortunately, it is not clear if all of the `_s` alternatives are thread safe. `strerror_s` is thread safe according to [its documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s?view=vs-2019). The documentation for [`localtime_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s?view=vs-2019), [`asctime_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/asctime-s-wasctime-s?view=vs-2019), and [`gmtime_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/gmtime-s-gmtime32-s-gmtime64-s?view=vs-2019) don't mention threads at all. Despite this, from what I can tell, they are. [`localtime_r`](https://en.cppreference.com/w/c/chrono/localtime), [`asctime_r`](https://en.cppreference.com/w/c/chrono/asctime), and [`gmtime_r`](https://en.cppreference.com/w/c/chrono/gmtime) will all be included in C2x (At least according to [en.cppreference.com](https://en.cppreference.com/w/)), however their descriptions seem to match their `_s` alternatives without the Annex K features. Despite the `_r` time functions being included in C2x I decided it would be better to use their `_s` counterparts. The addition of the `_r` is recent enough as to where my tdm64 gcc instillation does not support those functions. In addition to that, MSVC probably won't support them for another couple of centuries.

Going back to Annex K, there are no Annex K checks despite the usage of Annex K functions (`_s` functions). This is due to the fact tdm64 gcc and MSVC don't seem to properly declare macros related to Annex K. If Annex K is supported, `__STDC_LIB_EXT1__` should be defined. Yet, neither my tdm64 gcc nor my MSVC declared the macro. As such, the only Annex K check is the Windows check. It should still be mostly fine: the Windows `_s` functions are apart of the Windows API separate from Annex K.

There was an issue with the Windows API which caused the windowsapi.h file to be required. The issue is covered in that file.

Also, I added `/` as an alternative directory separator on Windows. Windows supports both `\` and `/` as separators in the Windows API, as such allowing the programmer to use `/` when using Path does no harm (Although I am not sure how long the `/` support has been around).

> You can use either forward slashes (/) or backslashes (\\) as the directory separators in a path. - https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019

Lastly, the GET_ERRNO macro had to be replaced by a function. The macro was a block which resolves to a value, a non-standard extension.

A few other changes are dotted about (Such as casting changes) which were required to quell the compiler warnings.


Jason2605:
This looks awesome thank you so much for this! I will be able to test this on a range of devices later on today hopefully and let you know how far i get through. Some points to note off the back of this - we should probably look to remove the VLAs within the interpreter anyways in all honesty however i understand your reasoning for not removing them as part of this PR (same with the strptime functionality).

Something that we will have to change eventually is the build instructions for the interpreter as we migrate to CMake and eventually remove the old makefile (can't really see a reason to keep it once), both within the README and on the documentation website. It will also be good to setup the CI/CD to also run on windows systems (that's for another task though, more of a self reminder).

TheGreatCookieMachine:
>I will be able to test this on a range of devices later on today hopefully and let you know how far i get through.

Awesome, thank you. Hopefully it goes well.

>...we should probably look to remove the VLAs within the interpreter anyways in all honesty...

Would you like me to remove the VLAs today? I should have free time later today.

Jason2605:
Yeah that would be nice! I don't mind if you want to hold off and do it as a separate task as well, i'll leave that up to you.

Jason2605:
So i tested this new CMake file on both Mac OS and Windows using CLion, and it seemed to work great, tests are passing on both which is awesome! I did however have to change the CMake version from `3.18.0` to `3.16.5` for CLion to play nice, I assume this is not an issue?

TheGreatCookieMachine:
> I don't mind if you want to hold off and do it as a separate task as well, i'll leave that up to you.

Alright. I will go ahead and do it in the same PR then. I should have it done in the next 4 hours or so.

>So i tested this new CMake file on both Mac OS and Windows using CLion, and it seemed to work great, tests are passing on both which is awesome!

Great! Thank you for testing it.

>I did however have to change the CMake version from 3.18.0 to 3.16.5 for CLion to play nice, I assume this is not an issue?

No, it isn't. The version listed in CMakeLists is actually the "minimum verified version". Essentially, it is the lowest version of CMake the CMakeLists is known to work on. Since my version of CMake is 3.18.0, that was the minimum version I could verify it ran on.

Jason2605:
Thank you so much for all the work that has gone into this PR! I'll get this merged into develop and work on getting CI/CD setup for windows. I'll also need to create a ticket for implementing `strptime` in windows along with probably an individual ticket for a round of testing before a release is created.

Issue #283, opened 2020-10-12
Update CI/CD to include windows


Jason2605:
# Update CI/CD to include windows
As of #278 Dictu can now be built on windows machines. We should adds windows OS to GitHub actions to ensure the test suite passes on windows machines as well.

Anu9518:
I would like to take this up

Issue #284, opened 2020-10-12
Update build instructions


Jason2605:
# Update build instructions
As of #278 CMake was introduced as a method for building Dictu. The README currently only describes how to build using the older makefile, we should extend the README (as well as the site documentation) to show how to build Dictu with CMake as well.

TheGreatCookieMachine:
(All terminal commands are shown assuming the current directory is Dictu's.)

## Author's Note

None of what I wrote here is intended to be put into the Readme or documentation, rather it is meant to be an accessible demonstration of the CMake CLI in the context of Dictu for those who wish to write the documentation. A lot of what is written here is written in more detail than is necessary.

This also ended up being way longer than I intended it to be. Sorry about that.

## Prologue

Building using CMake can be somewhat odd due to the fact that CMake is not a build file but rather a build file generator. This means its a two steps process: generate the build file and then use the build file. It is also further complicated by the fact that the CLI is not the only way to use CMake. On Windows, there is also CMake-GUI. Many IDEs also have support for CMake, with interaction methods differing from the CLI. Personally, I use VSCode with CMake-Tools. However, for simplicities sake I will just focus on the CLI.

## Generating the build file

### Basic Building

The first step is generating build files. To do so is fairly simple:

```
cmake ./
```

However, that clutters the main folder. If you want to send the new files into a build folder you can do this:

```
cmake -S ./ -B ./build
```

I wholeheartedly recommend doing it this way. Depending of the chosen generator, there may be numerous files and folders which are generated. It also makes it more difficult to remove the generated files if needed. Rather than just deleting the build folder, you have to pick and choose individual files.

### Options

To disable HTTPS and Linenoise, you do this:

```
<Previously mentioned setup> -D DISABLE_HTTP=TRUE -D DISABLE_LINENOISE=TRUE
```

If you want to ensure it is compiled as debug:

```
<Previously mentioned setup> -D CMAKE_BUILD_TYPE=DEBUG
```

Or if you want it to be release:

```
<Previously mentioned setup> -D CMAKE_BUILD_TYPE=RELEASE
```

### Generators

If you want to specify which generator file is generated, you can do so like so:

```
<Previously mentioned setup> -G "<generator>"
```

Where `<generator>` is a valid generator listed here: https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#manual:cmake-generators(7)

### CMakeCache

[...]

PR #288, opened 2020-10-14
#283 Update CI/CD to update test in windows

Anu9518:
type: feature
Subject: Modified main.yml to add test for windows os.


Jason2605:
Thanks for the PR but unfortunately it's not quite what i was looking for. As of #278 we now have the ability to generate our build files with CMake (more information here #284).

Jason2605:
Appreciate the effort with this, however i'm going to close this in favour of #290

PR #293, opened 2020-10-22
Implemented sin, cos, tan in math package.

MannarAmuthan:
Previous commit  adb786eda9a1ca5fe33676a02ff5d74a8f04f77f, that was failing because of wrong assertions. Those were fixed in this latest commit.

### DIFF
--- a/c/optionals/math.c
+++ b/c/optionals/math.c
@@ -179,6 +179,48 @@ static Value sqrtNative(VM *vm, int argCount, Value *args) {
     return NUMBER_VAL(sqrt(AS_NUMBER(args[0])));
 }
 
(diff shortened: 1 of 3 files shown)


Jason2605:
Appreciate the pull request! A small note is there was no documentation added about these 3 new methods, it would be great if we could add some to [this file](https://github.com/dictu-lang/Dictu/blob/develop/docs/docs/math.md) with a specific mention that these functions are in radians! Thank you.

MannarAmuthan:
As per suggestions I have added it in docs

Jason2605:
Perfect, thank you!

Issue #277, opened 2020-10-04
Write some more example Dictu programs


Jason2605:
# Summary
It would be good if we could get some more examples of Dictu programs to showcase a variety of features available in the language. This will also serve as a good task for testing the language in more "real" scenarios potentially uncovering unfound issues.

Vyvy-vi:
I can work on this.
Though, I might need a day or two to get a hang of Dictu

Vyvy-vi:
how many examples should I add?
Just the basic-understanding ones
or as many as I can?

Jason2605:
There is absolutely no limit, so feel free to do 1 or however many as you can!

WillDaSilva:
I would like to add a few

Jason2605:
Feel free to open a pull request @WillDaSilva 😄

WillDaSilva:
@Jason2605 will do. Is there any way to contribute modules that are written in Dictu to the standard library? If so, I have a few suggestions for modules that I would like to add.

Jason2605:
@WillDaSilva I generally avoid writing methods in Dictu for the standard library and opt for a C implementation wherever possible due to the order of magnitude in speed difference, however, I'm open to suggestions - what were you thinking of?

WillDaSilva:
@Jason2605 Is there a Discord server for Dictu? Or some other place where we can chat? This unrelated issue probably isn't the best place for it.

Jason2605:
@WillDaSilva There isn't no, but if you drop your username and discrim I'll send you a request and we can chat through it there

WillDaSilva:
Sounds good. I sent you an email with it, to avoid posting it publicly.

briandowns:
#500 adds another example.

briandowns:
#501 adds another example.

briandowns:
#504 adds another example.

PR #295, opened 2020-10-26
feat(examples): add isPalindrome.du

TGlide:
Addresses issue #277


Jason2605:
Nice example, thank you!

PR #299, opened 2020-10-27
Added Example for Design pattern - Factory method

Ajoe-T:
Hi, I have added factory method design pattern . Happy coding


Jason2605:
Looks nice, thanks for another example

PR #301, opened 2020-10-28
add bubble sort example

TGlide:
Addresses #277


Jason2605:
Thank you for another contribution! If you wanted to get fancy there are actually two optimisations you could add to this implementation, but i'll leave that up to you 😉.

TGlide:
Oh true! I'll make another PR then :)

PR #304, opened 2020-10-31
feat(examples): add binarySearchTree.du

siddharths2710:
Adding example for insert and search operations of Binary Search Tree.
Addresses issue #277


Jason2605:
Nice example, thank you!

Probably some stylistic changes i would make, but excusable since there aren't any style guides created for Dictu at this moment.

PR #305, opened 2020-10-31
Added docker installation of Dictu

avinashupadhya99:
- Added the ubuntu image of Dictu
- Will open an issue for the alpine image


Jason2605:
Nice addition, thank you!

PR #311, opened 2020-11-01
Added Dockerfile for alpine dictu

avinashupadhya99:
- Added the Dockerfile for *alpine* as the base image and compiled dictu with HTTP.
- Updated the Docker docs for the same.


Jason2605:
Thank you!

PR #312, opened 2020-11-01
added an example called the staircase

Pey-crypto:
Staircase example

### DIFF
--- a/examples/staircase.du
+++ b/examples/staircase.du
@@ -1,0 +1,22 @@
+
+    var no,spaces,rev;
+    spaces = input("Enter a number:").toNumber();
+    no = spaces;
+    rev = 0;
+
+    while(no>0){
+        spaces--;
+        rev++;
+        var spaces1=spaces;
+        while(spaces1>0){
+            print(" ");
+            spaces1--;
+        }
+        var rev1 = rev;
+        while(rev1>0){
+            print("#");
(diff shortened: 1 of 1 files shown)


Jason2605:
Unfortunately this code snippet won't run. The `++` and `--` operators are only in Dictu in a prefix format, not postfix. After fixing these errors i'm not entirely sure what the output was supposed to give me either.

Pey-crypto:
The program is supposed to generate a staircase
Given an input of 4 
      #
    ##
  ###
####
it would generate the above figure

Jason2605:
Unfortunately when you run the program that is not the outcome, the following is:

```
Enter a number:5
' '
' '
' '
' '
'1'
'
'
' '
' '
' '
'1'
'1'
'
'
' '
' '
'1'
'1'
'1'
'
'
' '
'1'
'1'
'1'
'1'
'
'
'1'
'1'
'1'
'1'
'1'
'
'
```

Pey-crypto:
Let me see if I can fix that

Jason2605:
Any update @Pey-crypto?

Pey-crypto:
yeah, I need to include another print method, like printf in c. Working on it

Jason2605:
I'm going to close this out due to inactivity. Feel free to re-open this if you're still working on it!

PR #328, opened 2020-11-15
Add Native Binding of VM Values

Kayomn:
Adds the ability for the host application to mark values that exist in the virtual machine as "bound" - marking them as being a dependency in native code.

This feature is very similar to Lua's reference registry and the `luaL_ref` / `luaL_unref` functions.

Effectively, three new user-facing functions are introduced:
  * `bindRef` -> Binds a value to native code, returning a unique 32-bit identifier to access it. In the scenario that the value is an object, this function will also remove said value from the GC's scope until an accompanying `unbindRef` returns it to the virtual machine.

  * `pushRef` -> Pushes the value at the unique identifier to the stack, pushing a nil value if the identifier is invalid. Regardless of whether the identifier is valid or invalid, some value will always be pushed to the stack.

  * `unbindRef` -> Unbinds the value from native code, returning full lifetime control to the virtual machine.

Alongside changes to the actual code, I've also added the `.vscode` metadata directory to the project `.gitignore`, as otherwise, people who use Visual Studio Code may accidentally submit unnecessary crap.

This feature has been tested by adding temporary code to VM initialization function, as the project cannot be built as a linkable library for integration into another application. The aforementioned temporary code has been removed, but of course this testing approach is no surefire and if you have any better means of testing the feature to suggest I'd be happy to use them.


Jason2605:
Thanks a bunch for the PR, however a few things i've found while testing. Similar to you, i added temporary code into initVM which i think is worth including

```c
...

    ObjString *string = copyString(vm, "test", 4);
    uint32_t ref = bindRef(vm, OBJ_VAL(string));

    string = copyString(vm, "test1", 5);
    uint32_t ref1 = bindRef(vm, OBJ_VAL(string));

    string = copyString(vm, "test2", 5);
    uint32_t ref2 = bindRef(vm, OBJ_VAL(string));

    string = copyString(vm, "test3", 5);
    uint32_t ref3 = bindRef(vm, OBJ_VAL(string));

//    printValue(OBJ_VAL(string));
//    printf("\n");
//
//    printValue(OBJ_VAL(string));
//    printf("\n");
//
//    printValue(OBJ_VAL(string));
//    printf("\n");

    unbindRef(vm, ref);
    unbindRef(vm, ref1);
    unbindRef(vm, ref2);
    unbindRef(vm, ref3);

    return vm;
```

- `freeRegistry` is never actually called, an apt place for this would be in `[freeVM](https://github.com/dictu-lang/Dictu/blob/develop/c/vm.c#L132)`.
- When freeing the registry we want to pass the capacity size rather than capacity size + 1.
- Given my testing above with binding string references, it seems as when binding 3 values the registry count reaches 3 while the capacity is still 2
- Similar when attempting to bind 4 values to the registry a segfault occurs (see below)

Screenshot 2020-11-15 at 18 21 23

PR #410, opened 2021-07-04
Added Switch Statement

manoharkakumani:
In Dictu Switch Statement we can even pass the string. But Switch statement only executes the first occurrence of the respective case block.


manoharkakumani:
@Jason2605  Errors are from the top branch itself. Even I got the same even I got same please find the attachment.
[test.txt](https://github.com/dictu-lang/Dictu/files/6759866/test.txt)

Screenshot from 2021-07-04 17-55-40

src/vm/compiler.c
```diff
@@ -1445,7 +1445,8 @@ ParseRule rules[] = {
         {NULL,     NULL,      PREC_NONE},               // TOKEN_IF
         {NULL,     and_,      PREC_AND},                // TOKEN_AND
         {NULL,     NULL,      PREC_NONE},               // TOKEN_ELSE
-        {NULL,     or_,       PREC_OR},                 // TOKEN_OR
+        {NULL,     or_,       PREC_OR},                 // TOKEN_OR               
+        {NULL,	   NULL,      PREC_NONE},               //TOKEN_MATCH
```
Jason2605, line 1449: The errors will be stemming from the fact you've added two new tokens, however, here you've only added one of the tokens to the array. The order of the enum is what is used to index on this array, so even though there's no logic required in here for `TOKEN_CASE` it still needs to be added here.

Hopefully that makes sense!

```suggestion
        {NULL,	   NULL,      PREC_NONE},               // TOKEN_MATCH 
        {NULL,	   NULL,      PREC_NONE},               // TOKEN_CASE
```

src/vm/vm.c
```diff
@@ -1361,7 +1360,12 @@ static DictuInterpretResult run(DictuVM *vm) {
             ip += offset;
             DISPATCH();
         }
-
+        CASE_CODE(COMPARE_JUMP):{
+            uint16_t offset = READ_SHORT();
+            Value a=pop(vm);
+            if (isFalsey(BOOL_VAL(valuesEqual(peek(vm,0), a)))) ip += offset;
```
Jason2605, line 1366: Here we can just use the value of `valuesEqual` since it returns a boolean value.

```suggestion
            if (!valuesEqual(peek(vm,0), a) ip += offset;
```

src/vm/compiler.c
```diff
@@ -2073,7 +2073,22 @@ static void ifStatement(Compiler *compiler) {
...
+    consume(compiler, TOKEN_RIGHT_PAREN, "Expect ')' after expression.");
+    consume(compiler, TOKEN_LEFT_BRACE, "Expect '{' before match body.");
+    consume(compiler, TOKEN_CASE, "Expect atleast one 'case' block.");
+      do{
+      expression(compiler);
+      int compareJump = emitJump(compiler, OP_COMPARE_JUMP);
+      consume(compiler, TOKEN_COLON, "Expect ':' after expression.");
+      statement(compiler);
+      patchJump(compiler,compareJump);
+      }while(match(compiler, TOKEN_CASE));
```
Jason2605, line 2088: Small formatting suggestions.

```suggestion
    do {
        expression(compiler);
        int compareJump = emitJump(compiler, OP_COMPARE_JUMP);
        consume(compiler, TOKEN_COLON, "Expect ':' after expression.");
        statement(compiler);
        patchJump(compiler, compareJump);
    } while(match(compiler, TOKEN_CASE)); 
```

src/vm/scanner.h
```diff
@@ -35,7 +35,7 @@ typedef enum {
     TOKEN_CLASS, TOKEN_ABSTRACT, TOKEN_TRAIT, TOKEN_USE, TOKEN_STATIC,
     TOKEN_PRIVATE, TOKEN_THIS, TOKEN_SUPER, TOKEN_DEF, TOKEN_AS,
     TOKEN_ENUM,
-    TOKEN_IF, TOKEN_AND, TOKEN_ELSE, TOKEN_OR,
+    TOKEN_IF, TOKEN_AND, TOKEN_ELSE, TOKEN_OR, TOKEN_MATCH,TOKEN_CASE,
```
Jason2605, line 38: Small formatting change

```suggestion
    TOKEN_IF, TOKEN_AND, TOKEN_ELSE, TOKEN_OR, TOKEN_MATCH ,TOKEN_CASE,
```

examples/match.du
```diff
@@ -0,0 +1,45 @@
+match(4){
```
Jason2605, line 1: Appreciate the example, would be awesome if we could work together on getting some test cases written along with some documentation!

Jason2605 (changes requested):
Thank you very much for the PR!

It would be interesting to get your thoughts on the use of "match" over "switch". I see in your opening PR you say it's because theres a greater range of data types available to use, but that doesn't necessarily stop us from using "switch". My only worry with the use of "match" is that it brings the connotation of powerful pattern matching, where this implementation is a lot closer to the switch statement. What do you think?

Jason2605:
Ahhh I also see another issue, since `match` is now a keyword, it's clashing with the `.match` function, this is why we're getting a ton of errors!

Maybe this is more of a reason to move to "switch", what do you think? 

See: https://dictu-lang.com/docs/error-handling/#matchfunc-success-func-error

manoharkakumani:
ok I'll change the keyword match to switch

manoharkakumani:
Updated keyword match to switch and working fine 
Screenshot from 2021-07-04 19-25-42

Jason2605:
This looks really nice, thanks!

Before merging we do need to get some documentation for the switch statement added along with a bunch of tests to ensure it works correctly and doesn't regress in the future.

Some points for future development. I wonder how easy it would be to have fall through cases, for example (maybe with special syntax to counter unwanted implicit fall through)

```
case 1:
case 2: {
    print("1 or 2!");
}
```
I do quite like not having to explicitly type `break;` to end the case block however as I've seen many a bug caused from an implicit fall through. 

Along with a `default` clause, matching up with the behaviour many other languages provide. 

```
switch (10) {
    case 1: {
        print("1!");
    }

    default: {
        print("Rest!");
    }
}
```

Jason2605:
Unfortunately upon testing some more there are some issues.

```
switch (1) {
   case 1: {
      var x = 10;
      print(x); // prints 1 not 10
   }
}
```

manoharkakumani:
> Unfortunately upon testing some more there are some issues.
> 
> ```
> switch (1) {
>    case 1: {
>       var x = 10;
>       print(x); // prints 1 not 10
>    }
> }
> ```
> 
> For implementation reference it may be worth checking [here](https://github.com/munificent/craftinginterpreters/blob/master/note/answers/chapter23_jumping/1.md).

manoharkakumani:
> Hi I resolved this.
> 
> ```
> switch (1) {
>    case 1: {
>       var x = 10;
>       print(x); // prints 1 not 10
>    }
> }
> ```
> I've implemented the default but instead of using the default, I implemented it as '$' what do you think do shell we proceed with '$' .
```
var x="hello";
switch (2) {
   case 1: {
      var x = 10;
      print(x); // prints 10
   }
      $: {
      var x = 10;
      var y = 11;
      print(x,y); // prints 10 and 11
   }
}
print(x);//prints hello
```

Jason2605:
> > Hi I resolved this.
> > ```
> > switch (1) {
> >    case 1: {
> >       var x = 10;
> >       print(x); // prints 1 not 10
> >    }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I've implemented the default keyword for the switch block.
> 
> ```
> var x="hello";
> switch (2) {
>    case 1: {
>       var x = 10;
>       print(x); // prints 10
>    }
>    default: {
>       var x = 10;
>       var y = 11;
>       print(x,y); // prints 10 and 11
>    }
> }
> print(x);//prints hello
> ```

Awesome thanks very much for fixing that!

> I've implemented the default keyword for the switch block.

This is great, thank you very much!! Now it's just a case of writing a few tests and some documentation for the new switch statement - I can work on doing that if you wish then we can get this merged in 😄.

manoharkakumani:
We will do some more checks before merge. I'm planning to implement fall keyword which allows us to excute the next case in switch block. Like
```
var x="hello";
switch (1) {
   case 1: {
      var x = 10;
      print(x); // prints 10
     fall;
   }
case (2):{
print ("two"); /*prints 2 as fall is in above case this will also execute*/
}
   default: {
      var x = 10;
      var y = 11;
      print(x,y); 
   }
}
print(x);//prints hello
```

Jason2605:
> We will do some more checks before merge. I'm planning to implement fall keyword which allows us to excute the next case in switch block. Like
> 
> ```
> var x="hello";
> switch (1) {
>    case 1: {
>       var x = 10;
>       print(x); // prints 10
>       fall;
>    }
> case (2):{
>         print ("two"); /*prints 2 
>         as fall is in above case this will also execute*/
> }
>  default: {
>       var x = 10;
>       var y = 11;
>       print(x,y); 
>    }
> }
> print(x);//prints hello
> ```
> 
> But right now I don't find free time. If I find free time I'll implement the above

That's fine, I appreciate the PR you've already done. What we can do is polish this one up with the documentation along with testing and implement the `fall` keyword (nice idea) in a new PR sometime in the future!

l-Luna:
syntax suggestion for fall through: maybe use `case value1, value2, value3:` (comma-seperated values) for cases that match multiple values, where you'd otherwise have a lone `fall` statement? (sorry if this was already considered)

briandowns:
Is there any technical reason to have the `:` and the `{` in the case syntax? It might be cleaner and preferred to use either one.

Jason2605:
> Is there any technical reason to have the `:` and the `{` in the case syntax? It might be cleaner and preferred to use either one.

We don't _have_ to have the `:` but what it does it separate the case expression from the statement. As for the braces, it's technically allowing any statement, so you don't theoretically need braces and can just place a single statement within a case (similar to an if statement without braces). If you want a block then you'll have to put braces (like an if statement with braces). Coming full circle, the `:` will allow clean separation between the case expression and the case statement (especially if the statement isn't a block).

Hope that all makes sense, let me know if not!

manoharkakumani:
Hello @Jason2605 
   I've implemented case value1, value2, value3: (comma-separated values)  as suggested by @l-Luna and inspired by Swift programming without using fallthrough. Even i felt like why to add new keyword 'fall'. Kindly do some checks from your end if every thing looks fine we can get this merged in. please find example below
```
switch("five"){
case "one":{
   print(1);
}
case "two":{
   print(2);
}
case "three":{
   print(3);
}
case "four":{
   print(4);
}
case "five", "six":{
   print("5 or 6");
}
}
//output 5 or 6
```

l-Luna:
> > Is there any technical reason to have the `:` and the `{` in the case syntax? It might be cleaner and preferred to use either one.
> 
> We don't _have_ to have the `:` but what it does it separate the case expression from the statement. As for the braces, it's technically allowing any statement, so you don't theoretically need braces and can just place a single statement within a case (similar to an if statement without braces). If you want a block then you'll have to put braces (like an if statement with braces). Coming full circle, the `:` will allow clean separation between the case expression and the case statement (especially if the statement isn't a block).
> 
> Hope that all makes sense, let me know if not!

if blocks are not valid expressions, could `case "label" { ... }` also be considered valid? as in, using colons for single statements and blocks for multi-line statements (or both if the user wanted)? IMO it'd look a lot cleaner but it's not rlly necessary

manoharkakumani:
image
@Jason2605 can you comment on this code snippet in statement function of compiler.c which part of dictu is being using it.

manoharkakumani:
> Hello @Jason2605 
>               I've implemented case value1, value2, value3: (comma-separated values)  as suggested by @l-Luna and inspired by Swift programming without using fallthrough. Kindly do some checks from your end if every thing looks fine we can get this merged in. please find example below
> ```
> switch("five"){
> case "one":{
>    print(1);
> }
> case "two":{
>    print(2);
> }
> case "three":{
>    print(3);
> }
> case "four":{
>    print(4);
> }
> case "five", "six":{
>    print("5 or 6");
> }
> }
> //output 5 or 6
> ```

@Jason2605 i feel implementation of 'fall' is necessary as in some cases fall is required
like if only second case is required to execute but for top case both need to be executed. I'll implement in future and request new PR. And I feel not to remove colon for both block statement and single statement as it looks clean to me, if you think it is unnessery we will look into it and statement function in compiler.c need come changes.

Jason2605:
It seems the multi-case functionality isn't quite working as expected unfortunately:

```c
switch (10) {
    case 1, 2: {
        print("1!");
    }

    case 10: {
        print("10!");
    }
}
```
Output
```
➜  Dictu git:(mano-develop) ./dictu test.du
➜  Dictu git:(mano-develop) 
```
(No print statement)

manoharkakumani:
> It seems the multi-case functionality isn't quite working as expected unfortunately:
> 
> ```c
> switch (10) {
>     case 1, 2: {
>         print("1!");
>     }
> 
>     case 10: {
>         print("10!");
>     }
> }
> ```
> 
> Output
> 
> ```
> ➜  Dictu git:(mano-develop) ./dictu test.du
> ➜  Dictu git:(mano-develop) 
> ```
> 
> (No print statement)

I've fixed the issue 
image

Jason2605:
Really sorry for the slow response rate on this @manoharkakumani, I think we're nearly there though! I've added a few test cases and some documentation now.

Jason2605:
Sorry for the absolute age this has taken @manoharkakumani !

PR #426, opened 2021-10-01
Update fibonacci example

WillDaSilva:
This update shows off a couple additional (more efficient) ways to compute the n-th fibonacci number.


Jason2605:
Thank you!

PR #428, opened 2021-10-01
Add `Math.gcd` and `Math.lcm`

### LINKED ISSUE #427: Add gcd and lcm to Math module
Both of these functions should take in 2 or more numbers, or a list of 2 or more numbers, and return the gcd or lcm of those numbers respectively.

The functions will need to be documented and tested.

WillDaSilva:
Closes #427

### DIFF
--- a/src/optionals/math.c
+++ b/src/optionals/math.c
@@ -267,6 +267,46 @@ static Value gcdNative(DictuVM *vm, int argCount, Value *args) {
     return NUMBER_VAL(result);
 }
 
+long long lcm(long long a, long long b) {
+    return (a * b) / gcd(a, b);
+}
+
+static Value lcmNative(DictuVM *vm, int argCount, Value *args) {
+    if (argCount == 1 && IS_LIST(args[0])) {
+        ObjList *list = AS_LIST(args[0]);
+        argCount = list->values.count;
+        args = list->values.values;
+    }
+
+    if (argCount < 2) {
+        runtimeError(vm, "lcm() requires 2 or more argument (%d given).", argCount);
+        return EMPTY_VAL;
+    }
+
+    for (int i = 0; i < argCount; ++i)
+        if (!IS_NUMBER(args[i])) {
+            runtimeError(vm, "A non-number value passed to lcm() as argument #%d", i);
+            return EMPTY_VAL;
+        }
+
+    double as_doubles[argCount];
+    for (int i = 0; i < argCount; ++i) {
+        as_doubles[i] = AS_NUMBER(args[i]);
+        if (fabs(round(as_doubles[i]) - as_doubles[i]) > FLOAT_TOLERANCE) {
+            runtimeError(vm, "A non-discrete number (%f) passed to lcm() as argument #%d", as_doubles[i], i);
+            return EMPTY_VAL;
+        }
+    }
+
+    long long as_longlongs[argCount];
+    for (int i = 0; i < argCount; ++i) as_longlongs[i] = round(as_doubles[i]);
+
+    long long result = as_longlongs[0];
+    for (int i = 1; i < argCount; ++i) result = lcm(result, as_longlongs[i]);
+
+    return NUMBER_VAL(result);
+}
+
 ObjModule *createMathsModule(DictuVM *vm) {
     ObjString *name = copyString(vm, "Math", 4);
     push(vm, OBJ_VAL(name));
@@ -289,7 +329,7 @@ ObjModule *createMathsModule(DictuVM *vm) {
     defineNative(vm, &module->values, "cos", cosNative);
     defineNative(vm, &module->values, "tan", tanNative);
     defineNative(vm, &module->values, "gcd", gcdNative);
-    // defineNative(vm, &module->values, "lcm", lcmNative);
+    defineNative(vm, &module->values, "lcm", lcmNative);
 
     /**
      * Define Math properties
--- a/src/optionals/math.c
+++ b/src/optionals/math.c
@@ -256,21 +256,24 @@ static Value gcdNative(DictuVM *vm, int argCount, Value *args) {
             return EMPTY_VAL;
         }
 
-    double as_doubles[argCount];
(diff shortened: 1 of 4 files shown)


src/optionals/math.h
```diff
@@ -7,6 +7,8 @@
 #include "optionals.h"
 #include "../vm/vm.h"
 
+#define FLOAT_TOLERANCE 0.00001
```
WillDaSilva, line 10: This value was decided arbitrarily. In the future if types are added for integers we can get rid of this and make the functions only accept integers. Until then using a float tolerance seems like a good idea, as otherwise people might enter numbers that appear to be whole numbers, but due to float math imprecision from previous calculations will actually not be.

src/optionals/math.c
```diff
@@ -221,6 +221,92 @@ static Value tanNative(DictuVM *vm, int argCount, Value *args) {
...
+
+static Value gcdNative(DictuVM *vm, int argCount, Value *args) {
+    if (argCount == 1 && IS_LIST(args[0])) {
+        ObjList *list = AS_LIST(args[0]);
+        argCount = list->values.count;
+        args = list->values.values;
+    }
+
+    if (argCount < 2) {
+        runtimeError(vm, "gcd() requires 2 or more argument (%d given).", argCount);
```
Jason2605, line 242:
```suggestion
        runtimeError(vm, "gcd() requires 2 or more arguments (%d given).", argCount);
```

src/optionals/math.c
```diff
@@ -221,6 +221,92 @@ static Value tanNative(DictuVM *vm, int argCount, Value *args) {
...
+        return EMPTY_VAL;
+    }
+
+    for (int i = 0; i < argCount; ++i)
+        if (!IS_NUMBER(args[i])) {
+            runtimeError(vm, "A non-number value passed to gcd() as argument #%d", i);
```
Jason2605, line 248: These will be 0-based. From a UX perspective for the error, I think 1-based would be a better solution. Do you agree?

src/optionals/math.c
```diff
@@ -221,6 +221,92 @@ static Value tanNative(DictuVM *vm, int argCount, Value *args) {
...
+
+static Value lcmNative(DictuVM *vm, int argCount, Value *args) {
+    if (argCount == 1 && IS_LIST(args[0])) {
+        ObjList *list = AS_LIST(args[0]);
+        argCount = list->values.count;
+        args = list->values.values;
+    }
+
+    if (argCount < 2) {
+        runtimeError(vm, "lcm() requires 2 or more argument (%d given).", argCount);
```
Jason2605, line 282:
```suggestion
        runtimeError(vm, "lcm() requires 2 or more arguments (%d given).", argCount);
```

src/optionals/math.c
```diff
@@ -221,6 +221,92 @@ static Value tanNative(DictuVM *vm, int argCount, Value *args) {
...
+        return EMPTY_VAL;
+    }
+
+    for (int i = 0; i < argCount; ++i)
+        if (!IS_NUMBER(args[i])) {
+            runtimeError(vm, "A non-number value passed to lcm() as argument #%d", i);
```
Jason2605, line 288: Same as above

src/optionals/math.c
```diff
@@ -221,6 +221,92 @@ static Value tanNative(DictuVM *vm, int argCount, Value *args) {
...
+    while (b > 0) {
+        r = a % b;
+        a = b;
+        b = r;
+    }
+    return a;
+}
+
+static Value gcdNative(DictuVM *vm, int argCount, Value *args) {
+    if (argCount == 1 && IS_LIST(args[0])) {
```
Jason2605, line 235: This is quite smart, however, I wonder how it'll play out.
```
>>> Math.lcm([2222]);
Error: File 'repl', line 1
       lcm() requires 2 or more argument (1 given).
```

I know that to determine the lowest common multiple we need at least two numbers, however, do you think it'll confuse people with the fact that it's using a list of length > 2 to contextually mean more than one argument? Even though we're technically only passing a single argument (the list). Would love to hear your thoughts on this!

Jason2605 (commented):
Looks good, just a few comments so far!

src/optionals/math.c
WillDaSilva: I thought about that, and was unsure. Since devs are the end-user, maybe they'd think "argument #1" is the second argument? I'm currently trying out a new way that should be less ambiguous, and make the error messages clear if a list is given.

WillDaSilva:
What are your thoughts on the code duplication between `gcdNative` and `lcmNative`? They could use a shared function, but abstracting that stuff out in C can be cumbersome (less we use a function macro, which I'm hesitant to do), so maybe it wouldn't be worth it if it was only for those two. Maybe there's some shared set of error checking / list unpacking cruft that could be abstracted away.

Jason2605:
> What are your thoughts on the code duplication between `gcdNative` and `lcmNative`? They could use a shared function, but abstracting that stuff out in C can be cumbersome (less we use a function macro, which I'm hesitant to do), so maybe it wouldn't be worth it if it was only for those two. Maybe there's some shared set of error checking / list unpacking cruft that could be abstracted away.

I'm not too bothered in its current state. I usually go for a rule of 3 before refactoring, if it gets unmanageable in the future we can look into ways, but for now it's fine.

Seems some of the tests are failing on windows machines unfortunately 😢

WillDaSilva:
> Seems some of the tests are failing on windows machines unfortunately 😢 

Looks like an error unrelated to the changes. Could you please try rerunning the workflow, if that's possible. Otherwise I can push an empty commit to trigger the workflow again.

Jason2605:
Appreciate you fixing the issues with MSVC. I've noticed in this PR you are also making use of `malloc`. Much like my comment of the other PR it would be great if we could try to use the macros available to us, see [memory.h for reference](https://github.com/dictu-lang/Dictu/blob/develop/src/vm/memory.h)

Jason2605:
Thank you!

PR #433, opened 2021-10-02
Add `Path.join`

### LINKED ISSUE #432: Add `Path.join`
This would be a function that accepts an arbitrary number of arguments (or a list with an arbitrary number of arguments), where all of the arguments are strings, and it would return those strings joined with the `Path.dirSeparator` in between them. It should handle the case were provided path segments end with `Path.dirSeparator`, e.g. `Path.join('/home/', 'will') == '/home/will'`.

WillDaSilva:
Closes #432

`Path.join` was added, along with tests and documentation.

### Context of the change :

#432

### Type of change :

- [X] New feature

### DIFF
--- a/src/optionals/path.c
+++ b/src/optionals/path.c
@@ -264,7 +264,7 @@ static Value joinNative(DictuVM *vm, int argCount, Value *args) {
     // separator, but having likely at most a few bytes of unused memory in the resulting string
     // object's char array should be fine.
 
-    char* str = calloc(resultSize, sizeof *str);
+    char* str = ALLOCATE(vm, char, resultSize);
     char* dest = str;
     for (int i = 0; i < argCount; ++i) {
         ObjString* src = AS_STRING(args[i]);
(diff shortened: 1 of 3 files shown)


src/optionals/path.c
```diff
@@ -233,6 +233,51 @@ static Value listDirNative(DictuVM *vm, int argCount, Value *args) {
...
+    }
+
+    // resultSize = # of dir separators that will be used + length of each string arg
+    size_t resultSize = abs(argCount - 1); // abs is needed her because of a clang bug
+    for (int i = 0; i < argCount; ++i) resultSize += AS_STRING(args[i])->length;
+    // It's possible for resultSize to be too large if the strings already end with the dir
+    // separator, but having likely at most a few bytes of unused memory in the resulting string
+    // object's char array should be fine.
+
+    char* str = calloc(resultSize, sizeof *str);
```
Jason2605, line 267: When allocating memory it's best to try and use the macros available to us. This will 1, cause a garbage collection to occur if required and 2. allow us to swap out the memory allocation scheme at a later date if ever required much easier.

```suggestion
    char* str = ALLOCATE(vm, char*, resultSize);
```

Jason2605:
Thank you!

PR #442, opened 2021-10-02
Class properties

Jason2605:

This PR changes class properties so that they are copied to the new class when inheriting. This means when we're getting a class property we can just check the instance class rather than walking the inheritance chain. This will come at a memory cost, however, will make getting class attributes faster in theory.

### Context of the change :

This was changed so that getting a class attribute can be checked within a single hash table rather than walking the inheritance chain.

### Type of change :

- [x] New feature

#


Jason2605:
This won't work on second thought. If we copy class variables at inheritance and change the class variable of a class down the chain this wouldn't be reflected.

PR #444, opened 2021-10-02
Fix memory issue with `Path.join`

WillDaSilva:

The tests weren't running for `Path.join` previously, so a small memory issue slipped in. This PR makes the `Path.join` tests run, and fixes that bug.

### Type of change :

- [X] Bug fix


Jason2605:
Thanks!

PR #445, opened 2021-10-02
Strip leading dir separators in `Path.join`

WillDaSilva:

Previously this was the behaviour of `Path.join` with leading directory separators:

```
>>> Path.join('/tmp/', '/abcd/', '/efg/');
/tmp//abcd//efg
```

Now this is the behaviour:

```
>>> Path.join('/tmp/', '/abcd/', '/efg/');
/tmp/abcd/efg
```

### Type of change :

- [X] Bug fix

### DIFF
--- a/src/optionals/path.c
+++ b/src/optionals/path.c
@@ -263,21 +263,50 @@ static Value joinNative(DictuVM *vm, int argCount, Value *args) {
     for (int i = 0; i < argCount; ++i) {
         part = AS_STRING(args[i]);
         resultSize += part->length;
-        resultSize -= i && part->chars[0] == DIR_SEPARATOR;
-        resultSize -= part->chars[part->length-1] == DIR_SEPARATOR;
+        // Account for leading DIR_SEPARATOR chars to be removed
+        for (int j = 0; j < part->length; ++j) {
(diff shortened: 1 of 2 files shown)


Jason2605:
For visibility I'll drop a note here.

As discussed it would be good if we could handle cases where the directory separator is occurring frequently

##### Current
```
>>> Path.join(['/tmp', '/////abcd/////', '/efg/']);
/tmp/////abcd/////efg
```

##### Suggested
```
>>> Path.join(['/tmp', '/////abcd/////', '/efg/']);
/tmp/abcd/efg
```

WillDaSilva:
> For visibility I'll drop a note here.
> 
> As discussed it would be good if we could handle cases where the directory separator is occurring frequently
> 
> ##### Current
> ```
> >>> Path.join(['/tmp', '/////abcd/////', '/efg/']);
> /tmp/////abcd/////efg
> ```
> 
> ##### Suggested
> ```
> >>> Path.join(['/tmp', '/////abcd/////', '/efg/']);
> /tmp/abcd/efg
> ```

This has been added.

Jason2605:
Thank you!

Issue #430, opened 2021-10-01
Add `Path.walk`


WillDaSilva:
I'd be happy if it had the same semantics as Python's [`os.walk`](https://docs.python.org/3/library/os.html#os.walk)

WillDaSilva:
This will have to be implemented using a class (e.g. `Path.Walker`), since Dictu doesn't have stateful functions aka generators. The reason it needs to be stateful is because doing the entire tree traversal before returning limits the usefulness. For instance, if you wanted to search from the root directory and this feature returned a list of every path it would take ages to execute, and the results would use a large amount of memory. Instead we can create an instance of the class given the root of the filesystem tree to be traversed, and have a `next` method (or some similar name) to get the next `[path, dirnames, filenames]` triplet. When you call the `next` method you can optionally provide an updated `dirnames` list to prune the traversal.

Jason2605:
Rationale makes perfect sense to me, I like it

Jason2605:
446 implements the functionality required within Dictu, however, I'll keep this issue open as a reminder to add it to the stdlib once we work out a viable way of running Dictu code.

PR #446, opened 2021-10-03
Add path walker example

WillDaSilva:

This PR adds an example that has a `Walker` class. It traverses a file system starting from a given directory, or the current directory if none is given (as in the example). After creating a `Walker` instance one can repeatedly call the `next` method to obtain lists of three elements: `[dirpath, dirnames, filenames]`.
- `dirpath` is the path to the directory it just searched
- `dirnames` is a list of the directories in `dirpath`; edits to this list can change the traversal, as shown in the example by the removal of the `.git` directory from the traversal
- `filenames` is a list of the files in `dirpath`

The `next` method returns `nil` once the traversal has finished.

Once we figure out how to use Dictu code in the standard library the `Walker` class in this example can be added to it as `Path.Walker`.

### Context of the change :

#430

### Type of change :

- [X] New feature


Jason2605:
I wonder if a nice change would be to pass a list of excluded directories to the `Walker` class when being instantiated and we have that logic as part of the class? I can see it being used enough where forcing people to implement it themselves may become cumbersome. What do you think?

Other than that though looks great!

WillDaSilva:
> I wonder if a nice change would be to pass a list of excluded directories to the `Walker` class when being instantiated and we have that logic as part of the class? I can see it being used enough where forcing people to implement it themselves may become cumbersome. What do you think?

Good idea. That would provide some nice convenience for the cases were the greater flexibility is not needed.

Jason2605:
Thank you!

Issue #450, opened 2021-10-03
[FEATURE] `Path.isLink`


WillDaSilva:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

The `Path` library does not have a method to check if a given path is symbolic link.

### Describe the solution you'd like

A `Path.isLink` method that takes a single string argument, the path, and returns a bool indicating whether it is or is not a symbolic link.

### Describe alternatives you've considered

Could also be called `Path.isSymLink`, which is more accurate, but more verbose.

Kishore-abhimanyu:
Can you assign me this issue. I would like to contribute to this?

Jason2605:
Done @Kishore-abhimanyu !

Kishore-abhimanyu:
@Jason2605 Can you share some test cases for this?

Jason2605:
For testing the method is working for non-symbolic links just use any directory within the project. For testing symbolic links we may have to get a bit creative to appease the tests.

You do have [System.platform](https://dictu-lang.com/docs/standard-lib/system/#constants) available to you so that you can build the tests in a way that are platform related checking for symlinks that are there by default. Will have to do some research into finding these.

WillDaSilva:
@Kishore-abhimanyu you'll have to write your own test cases. I would recommend creating a directory under `tests/path` that has a symlink to a regular file, a symlink to a directory, and a broken symlink. Note that git can commit symlinks. Then add the file `tests/path/isLink.du`, and use `assert` with `Path.isLink` to test against those symlinks and non-symlinks under `tests/path`.

Jason2605:
> @Kishore-abhimanyu you'll have to write your own test cases. I would recommend creating a directory under `tests/path` that has a symlink to a regular file, a symlink to a directory, and a broken symlink. Note that git can commit symlinks. Then add the file `tests/path/isLink.du`, and use `assert` with `Path.isLink` to test against those symlinks and non-symlinks under `tests/path`.

This may cause issues for windows machines. If i'm not mistaken (untested) a symlink created on say a unix machine and pulled down on a windows machine creates a text file rather than an actual symbolic link.

Kishore-abhimanyu:
will add the windows functionality also

liz3:
https://github.com/dictu-lang/Dictu/pull/754

PR #457, opened 2021-10-04
Added functionality for isPath

Kishore-abhimanyu:
I worked on the ispath to check whether the given string is symbolic link or not

### Context of the change :
 

https://github.com/dictu-lang/Dictu/issues/450#issue-1014437358

- [ ] Bug fix

- [x] New feature

### DIFF
--- a/docs/docs/standard-lib/path.md
+++ b/docs/docs/standard-lib/path.md
@@ -114,3 +114,13 @@ Path.join('/tmp', 'abcd', 'efg') == '/tmp/abcd/efg';
 Path.join(['/tmp', 'abcd', 'efg']) == '/tmp/abcd/efg';
 Path.join('/tmp/', 'abcd/', 'efg/') == '/tmp/abcd/efg';
 ```
+
+### Path.isPath(string)
+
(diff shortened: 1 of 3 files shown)


Kishore-abhimanyu:
> Why is the function called `isPath` if it's checking whether something is a symbolic link or not? Not all paths are symbolic links.

File name has been changed. Could you check now?

Jason2605:
The name the issue stated is fine, `isLink`.

Few things as well, you will need to add the test file to [here](https://github.com/dictu-lang/Dictu/blob/develop/tests/path/import.du).

The code seems to be slightly mis-indented unfortunately.

And finally, this will break windows builds. If we intend this to not work on windows (which may be a viable option to begin with) we either need to 1) remove this and make sure it doesn't get compiled on windows systems entirely. 2) Do something similar to Python where it just returns False on runtimes that don't support it (less hot on this as to me it feel like a silent fail).

WillDaSilva:
That new name also doesn't make sense. It should be called `isLink`.

Kishore-abhimanyu:
Could you review now?

PR #465, opened 2021-10-09
Added list.limit() utility method

MannarAmuthan:
Added New Feature: List limit method called limit(). To return a list with limited size. If the size is less than 0 , it will return empty list, and if the size is more than list's length, then entire list will be return.

Updated Docs , and tests are ran successfully .


Jason2605:
I'm not entirely sure how this is different to list slicing, sorry

https://dictu-lang.com/docs/collections/lists/#slicing

PR #482, opened 2021-12-24
add additional math mod values

briandowns:
Signed-off-by: Brian Downs <bdowns@Brians-MacBook-Pro.local>

Add additional native properties in math mod.

Resolves: 
 
#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

### DIFF
--- a/docs/docs/standard-lib/math.md
+++ b/docs/docs/standard-lib/math.md
@@ -31,6 +31,13 @@ import Math;
 |--------------|--------------------------------------------------------|
 | Math.pi      | The mathematical constant: 3.14159265358979            |
 | Math.e       | The mathematical constant: 2.71828182845905            |
+| Math.phi     | The mathematical constant: 1.61803398874989            |
+| Math.sqrt2   | The mathematical constant: 1.41421356237309            |
+| Math.sqrte   | The mathematical constant: 1.61803398874989            |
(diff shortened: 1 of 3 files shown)


Jason2605:
Hi there @briandowns!

Firstly thanks for the PR! Unfortunately this has broken the tests, would be good to extend the maths tests to include the new values you've added as well

briandowns:
Ugh, my fault. Updates pushed and I'll monitor to make sure CI passes. Thank you!

Jason2605:
One more test is using `Math.pi`!

Interesting point here though is that we are using the `double` type to represent all values within Dictu. This means numeric values will only have precision for ~15 decimal places, anything after that is lost (silently).

briandowns:
That makes sense. I was pulling this from an extremely modified fork. I'll update.

briandowns:
Thanks for the help on this @Jason2605 . I have a few more PRs to add for some more features if that's ok.

Jason2605:
No worries! Yeah I'm all for extra contributions, feel free!

Last thing we need to do here is to just add these new values to the docs, I probably should have said previously!

See here: https://github.com/dictu-lang/Dictu/blob/develop/docs/docs/standard-lib/math.md#constants

briandowns:
All set!

Jason2605:
Thank you for the PR!

PR #483, opened 2021-12-24
initial log module impl

briandowns:

Resolves: 
 
#

### Type of change:

Adds a simple log package.

I also have a structured logger written entirely in Dictu using the JSON module. Was hoping for feedback on this and possible integration of the structured one.

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

### DIFF
--- a/docs/docs/standard-lib/log.md
+++ b/docs/docs/standard-lib/log.md
@@ -1,0 +1,69 @@
+---
+layout: default
+title: Log
+nav_order: 16
+parent: Standard Library
+---
+
+# Log
+{: .no_toc }
+
+## Table of contents
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+## Log
+
+To make use of the Log module an import is required.
+
+```cs
+import Log;
+```
+
+### Constants
+
+| Constant   | Description                                                           |
+| ---------- | --------------------------------------------------------------------- |
+| Log.stdout | The default file descriptor where a process can write output.         |
+| Log.stderr | An input stream where data is sent to and read by a program.          |
+| Log.stdin  | The default file descriptor where a process can write error messages. |
+
+### Log.print(string)
+
+```cs
+Log.print("something extremely interesting");
+```
+
+### Log.println(string)
+
+```cs
+Log.println("hello, world! println");
+```
+
+### Log.new(number)
+
+Create a new instance of a logger.
+
+```cs
+var log = Log.new(Log.stdout).unwrap();
+```
+
+### log.setPrefix(string)
+
+A prefix can be set on the logger that will be included in the output just before the user provided content.
+
+```cs
+log.setPrefix("prefix");
+```
+
+### log.unsetPrefix(string)
+
+Remove the prefix on the logger. This is a noop if there was no prefix previously set.
+
+```cs
+log.unsetPrefix();
+```
--- a/docs/docs/standard-lib/log.md
+++ b/docs/docs/standard-lib/log.md
@@ -29,19 +29,19 @@ import Log;
 | Constant   | Description                                                           |
 | ---------- | --------------------------------------------------------------------- |
 | Log.stdout | The default file descriptor where a process can write output.         |
-| Log.stderr | An input stream where data is sent to and read by a program.          |
-| Log.stdin  | The default file descriptor where a process can write error messages. |
+| Log.stderr | The default file descriptor where a process can write error messages. |
+| Log.stdin  | An input stream where data is sent to and read by a program.          |
 
 ### Log.print(string)
 
 ```cs
-Log.print("something extremely interesting");
+Log.print("something extremely interesting"); 2021/11/27 11:49:14 something extremely interesting
 ```
 
 ### Log.println(string)
 
 ```cs
-Log.println("hello, world! println");
+Log.println("hello, world! println"); // 2021/11/27 11:49:14 hello, world! print
 ```
 
 ### Log.new(number)
--- a/docs/docs/standard-lib/log.md
+++ b/docs/docs/standard-lib/log.md
@@ -65,7 +65,7 @@ Log.fatalln("hello, world! fatalln"); // 2021/11/27 11:49:14 hello, world! print
 Create a new instance of a logger.
 
 ```cs
-var log = Log.new(Log.stdout).unwrap();
+const log = Log.new(Log.stdout).unwrap();
 ```
 
 ### log.setPrefix(string)
@@ -82,6 +82,18 @@ Remove the prefix on the logger. This is a noop if there was no prefix previousl
 
 ```cs
 log.unsetPrefix();
+```
+
+### log.print(string)
+
+```cs
+Log.print("something extremely interesting"); // 2021/11/27 11:49:14 something extremely interesting
+```
+
+### log.println(string)
+
+```cs
+Log.println("hello, world! println"); // 2021/11/27 11:49:14 hello, world! print
 ```
 
 ### log.fatal(string)
--- a/docs/docs/standard-lib/log.md
+++ b/docs/docs/standard-lib/log.md
@@ -35,13 +35,29 @@ import Log;
 ### Log.print(string)
 
 ```cs
-Log.print("something extremely interesting"); 2021/11/27 11:49:14 something extremely interesting
+Log.print("something extremely interesting"); // 2021/11/27 11:49:14 something extremely interesting
 ```
 
 ### Log.println(string)
 
 ```cs
 Log.println("hello, world! println"); // 2021/11/27 11:49:14 hello, world! print
+```
+
+### Log.fatal(string)
+
+Log the given output and exit the program with an exit code of 1.
+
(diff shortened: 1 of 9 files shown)


Jason2605:
Hi, thanks for the PR!

Couple of things, the tests here won't actually be running since the import file hasn't been added here: https://github.com/dictu-lang/Dictu/blob/develop/tests/runTests.du

The problem is, however, is that the fatallog will cause the tests to fail since it's exiting with a non-zero exit code.

> I also have a structured logger written entirely in Dictu using the JSON module. Was hoping for feedback on this and possible integration of the structured one.

I think having a logging module within the stdlib is definitely something I think is a good idea, it would be nice if we eventually end up with something like other languages give where you can choose the output stream (stdout, stderr, file, db etc) and something like log levels. That can always come at a later date though, and getting something in is a great start!

briandowns:
I left it incomplete more for feedback purposes. I didn't want to go much further if it wasn't a direction you were interested in. I'll get this a bit more clean and fleshed out and see what you think. Thanks!

Jason2605:
> I left it incomplete more for feedback purposes. I didn't want to go much further if it wasn't a direction you were interested in. I'll get this a bit more clean and fleshed out and see what you think. Thanks!

Makes sense! Yep all for a logging library 👍

briandowns:
Aside from the failing tests, I think this is ready to be looked at again.

src/optionals/log.c
```diff
@@ -0,0 +1,298 @@
...
+        ptm->tm_hour, ptm->tm_min, ptm->tm_sec, msg);
+
+    exit(1);
+}
+
+ObjAbstract* newLogObj(DictuVM *vm) {
+    ObjAbstract *abstract = newAbstract(vm, freeLog);
+    push(vm, OBJ_VAL(abstract));
+
+    Log *log = ALLOCATE(vm, Log, 1);
```
Jason2605, line 228:
```suggestion
    Log *log = ALLOCATE(vm, Log, 1);
    log->of = NULL;
    log->prefix = NULL;
```

We will want to initialise both of those to NULL. If a prefix is never set for example, when trying to free a Log object we're checking for NULL, we want to ensure that it's actually NULL rather than a garbage value.

src/optionals/log.c
```diff
@@ -0,0 +1,298 @@
...
+    return NIL_VAL;
+}
+
+static Value logObjUnsetPrefix(DictuVM *vm, int argCount, Value *args) {
+    if (argCount != 0) {
+        runtimeError(vm, "unsetPrefix() takes 0 argument (%d given)", argCount);
+        return EMPTY_VAL;
+    }
+
+    Log *log = AS_LOG(args[0]);
```
Jason2605, line 130: Will this leak memory? My initial feelings are we will need to free the old `log->prefix` if it's not NULL.

Jason2605 (changes requested):
Couple of comments but overall it looks great, thank you!

*Edit: I've noticed even with these changes it seems the testing output is not what it should be. I think it'll require some more investigation into why it's causing changes to alternative tests.

briandowns:
Thanks for the feedback. I'll get these changes made. As for testing, it would appear as though the UNIX timestamps are different in the strptime tests. I don't know why that'd be the case.

Jason2605:
image

Interestingly on my system this is the output from the tests. Obviously this isn't quite right as the process module tests should output, and there should be a final all clear message. What makes this slightly more odd too is that it's exiting with a 0 status code.

Github Actions output:
```
2021/11/26 14:57:09 print output to stderr
2021/11/26 14:57:09 hello, world! print
2021/11/26 14:57:09 hello, world! println
2021/11/26 14:57:09 PREFIX println output to stdout with prefix
2021/11/26 14:57:09 print output to stdout without prefix
ls: write error: Bad file descriptor
ls: write error: Bad file descriptor
ls: write error: Bad file descriptor
Error: File '/home/runner/work/Dictu/Dictu/tests/process/run.du', line 20
       assert() was false!

Error: Process completed with exit code 1.
```

I'll keep looking into this locally and see if I can see whats going on

Jason2605:
@briandowns I've pushed to this PR the following changes:
- I've moved the creation of `newResultSuccess(vm, OBJ_VAL(abstract));` to be wrapped around a push / pop to ensure that the abstract object is not cleaned up by the GC
- I've stopped the closing of files. This was closing stderr / stdout and when writing to it later on will end up leading to undefined behaviour.

I've noticed the tests are failing on windows as the constants are undefined, but at least there is progress!

briandowns:
Oh man, I really overlooked that. Thank you so much!

Jason2605:
> Oh man, I really overlooked that. Thank you so much!

It's no problem! Is this ready to be marked for review or are you still thinking of extra additions?

docs/docs/standard-lib/log.md
```diff
@@ -0,0 +1,69 @@
...
+import Log;
+```
+
+### Constants
+
+| Constant   | Description                                                           |
+| ---------- | --------------------------------------------------------------------- |
+| Log.stdout | The default file descriptor where a process can write output.         |
+| Log.stderr | An input stream where data is sent to and read by a program.          |
+| Log.stdin  | The default file descriptor where a process can write error messages. |
```
Jason2605, line 33:
```suggestion
| Log.stderr |  The default file descriptor where a process can write error messages.     |
| Log.stdin   | An input stream where data is sent to and read by a program.                    |
```

src/optionals/log.c
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    struct tm *ptm = timeNow();
+
+    char *msg = AS_CSTRING(args[0]);
+    printf("%04d/%02d/%02d %02d:%02d:%02d %s", 
+        ptm->tm_year+2000-100, ptm->tm_mon, ptm->tm_mday, 
+        ptm->tm_hour, ptm->tm_min, ptm->tm_sec, msg);
+
+    return newResultSuccess(vm, NIL_VAL);
```
Jason2605, line 42: This can't have a recoverable erroneous situation from what I can gather, so I think returning nil rather than a `Result` type makes more sense here

 ```suggestion
    return NIL_VAL;
```

File: src/optionals/log.c [hunk 5]
```diff
@@ -0,0 +1,302 @@
...
+    }
+    
+    struct tm *ptm = timeNow();
+
+    char *msg = AS_CSTRING(args[0]);
+    printf("%04d/%02d/%02d %02d:%02d:%02d %s\n", 
+        ptm->tm_year+2000-100, ptm->tm_mon, ptm->tm_mday, 
+        ptm->tm_hour, ptm->tm_min, ptm->tm_sec, msg);
+
+    return newResultSuccess(vm, NIL_VAL);
```
Jason2605, line 58: This can't have a recoverable erroneous situation from what I can gather, so I think returning nil rather than a Result type makes more sense here

```suggestion
    return NIL_VAL;
```

docs/docs/standard-lib/log.md
```diff
@@ -0,0 +1,69 @@
...
+### Log.print(string)
+
+```cs
+Log.print("something extremely interesting");
+```
+
+### Log.println(string)
+
+```cs
+Log.println("hello, world! println");
```
Jason2605, line 44: I think giving an example output may be useful!

```suggestion
Log.println("hello, world! println"); // 2021/11/27 18:37:48 hello, world! println
```

docs/docs/standard-lib/log.md
```diff
@@ -0,0 +1,69 @@
...
+```cs
+log.setPrefix("prefix");
+```
+
+### log.unsetPrefix(string)
+
+Remove the prefix on the logger. This is a noop if there was no prefix previously set.
+
+```cs
+log.unsetPrefix();
```
Jason2605, line 68: I wonder if you think this should also remove the prefixed timestamp that is appended to all logs (in case they want a custom time format for example)?

We can look at this as a future PR if you wish

Jason2605 (changes requested):
Few comments on this, but thanks for the PR!

An extra point too:
- The fatal / fatalln methods aren't in the docs

docs/docs/standard-lib/log.md
briandowns: This is something I had considered but wasn't too sure about. I had modeled this after Go for the most part. I definitely think there's a lot of room for improvement where we could allow a user to set default fields, add fields, ad hoc, set custom timestamp formats, etc. A separate PR might be best to start investigating those ideas. However I'm open to your suggestions around that.

Jason2605: Happy with investigating at a later date yep!

Jason2605:
I still need to add the win guards (shouldn’t take me too long) and we still need the fatal / fatalln entries to the docs, after that we should be all good!

briandowns:
> We still need the fatal / fatalln entries to the docs, after that we should be all good!
> 
> Edited* ignore the first part, I was reading your other PR too 😓

Docs updated!

Jason2605:
Perfect, thanks for this!

PR #484, opened 2021-12-25
add getAttributes to instances.c

briandowns:

Add function that will return all public attributes on an instance of a class.

Resolves: 
 
#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping


Jason2605:
This is functionally very similar to `.methods`, which led me to question myself. I think `.getAttributes` or `.attributes` is probably better suited name than `.methods`; this is due to the fact that `.getAttribute()` searches methods, public object attributes and class variables. I wonder if it's worth merging `.methods()` and `.getAttributes()` into a single method that (maybe?) returns a dict with listing for all three options

e.g
```
class Test {
    var x = 10;

    init(var y=20) {}

    test() {}
}

print(Test().getAttributes());
// Returns
/*
{
    "classVariables": ["x"],
    "properties": ["y"]
    "methods": ["init", "test"]
}
*/
```

What do you think?

briandowns:
All of that seems to make sense. I'll adjust and see what it looks like.

Jason2605:
> All of that seems to make sense. I'll adjust and see what it looks like.

Perfect! Let me know if i can help at all.

Jason2605:
I think once we sort out this PR we're probably good for a new release! There has been a lot added between 0.22.0 and now!

briandowns:
Sounds good to me! I'm working on this PR now. Once I get this one done, it'll support a very naive ORM I have in my fork which I'm excited to get ported over.

briandowns:
Might need a hand figuring this one out. @Jason2605 

```
Undefined property 'getAttributes'
```

briandowns:
Looks like `strptime.du` is up to no good.

Jason2605:
Thanks for this @briandowns! I've pushed a few changes and will jot down my reasoning here!

- I've made it so that the `classVariables` now walk the inheritance chain and copy all from previous superclasses. `classVariables` are not copied down the inheritance chain and so we need to ensure we walk it for those!
- I've removed the `push()` and `pop()` calls within each loop, the `instance` will still be on the stack so we don't need to push any instance values to the stack (as the GC will grey them all from the parent instance on the stack)
- I've made it so the `classVariable` list pulls values from `instance->klass->publicProperties` rather than `instance->publicFields` (and vice-versa for `properties`)
- Updated some of the surrounding `push()` / `pop()` calls too.

If any of this doesn't make sense let me know!

briandowns:
This looks great. Sorry for the extra work.

Jason2605:
> This looks great. Sorry for the extra work.

It's all good, thanks for the PR!

PR #485, opened 2021-12-26
update http module with requests, methods, and statuses

briandowns:

Resolves: 
 
#

### Type of change:

Adds access to common headers, response codes, and messages.

- [ ] Bug fix

- [x] New feature

#

### Housekeeping


briandowns:
I can update docs if this changes is deemed acceptable.

tests/http/status.du
```diff
@@ -0,0 +1,3 @@
+import HTTP;
+
+assert(HTTP.STATUS_OK == 200);
```
Jason2605, line 3: Remember to add this file to: https://github.com/dictu-lang/Dictu/blob/develop/tests/http/import.du

Other than that looks fine to me! I wonder how we can add all of these to the docs without making the page too unmanageable.

briandowns: I took a stab at some verbiage to describe the added functionality in the docs. If you want it very explicit, I'm happy to do so.

Jason2605: I'm pretty happy with that to be honest. I think listing out every constant on the page would be a bit much, thanks!

Issue #473, opened 2021-10-19
[FEATURE] Env.readFile()


Jason2605:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

I currently use a utility function to read env vars from an `.env` file, however, I feel this is functionality that will be used enough that it warrants something built-in to accomplish this.

### Describe the solution you'd like

```
import Env;

Env.readFile(".env");
```

Maybe even the default argument is ".env"
```
import Env;

Env.readFile(); // Implicitly reads ".env"
```

### Describe alternatives you've considered

```
import Env;

def readEnv(path) {
    with(path, "r") {
        var line;
        // When you reach the end of the file, nil is returned
        while((line = file.readLine()) != nil) {
            if (!line)
                continue;

            var [variable, value] = line.split("=");
            Env.set(variable, value);
        }
    }
}
```

### Additional context

The above implementation is a bit naive and could do with some work, for example, if an env var value contains a `=` it's going to cause issues.

briandowns:
I've been using this in a few projects and it's been working well so far.

```
import Env;
import System;

const ENV_FILE_NAME = ".env";
const COMMENT_CHAR = "#";
const SEPERATOR = "=";

/**
 * load loads all entries in the local .env unless
 * a path is provided and then that file is loaded.
 */
def load(path) {
    var full_path = "";

    if (path == "") {
        full_path = ENV_FILE_NAME;
    } else {
        full_path = path;
    }

    var data = {};

    with(full_path, "r") {
        var line;
        var line_num = 0;

        while((line = file.readLine()) != nil) {
            line_num += 1;

            if (line.startsWith(COMMENT_CHAR)) {
                continue;
            }

            if (line.contains(SEPERATOR)) {
                var comp_count = line.split(SEPERATOR).len();
                if (comp_count == 2) {
                    var key = line.split(SEPERATOR)[0].strip();
                    var value = line.split(SEPERATOR)[1].strip();
                    
                    data[key] = value;
                } else {
                    return Error("error: line {} malformed entry".format(line_num)); 
                }
            }
        }
    }

    if (data.len() == 0) {
        return success(1);
    }

    var keys = data.keys();
    for (var i = 0; i < keys.len(); i += 1) {
        Env.set(keys[i], data[keys[i]]);
    }

    return Success(0);
}

// main
{
    var res = load("").match(
        def (result) => result,
        def (error) => {
            print(error);
            System.exit(1);
        }
    );

[...]

PR #489, opened 2021-12-26
Add optional argument to split

Jason2605:

This PR adds a new optional argument to `str.split()` which determines the amount of times a string will actually be broken up by the delimiter. This allows us to keep the delimiter within the string after a certain amount of occurrences have already been found.

E.g
```
"Dictu is awesome!".split(" ", 0); // ['Dictu is awesome!']
"Dictu is awesome!".split(" ", 1); // ['Dictu', 'is awesome!']
"Dictu is awesome!".split(" ", 2); // ['Dictu', 'is', 'awesome!']
```

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

### Note
This may come in useful with regards to #473

### DIFF
--- a/docs/docs/strings.md
+++ b/docs/docs/strings.md
@@ -167,9 +167,11 @@ Returns true if a string ends with a given string.
 Returns a list of strings, split based on a given delimiter.
 Returns a list of all characters in a string if an empty string is passed as delimiter.
 
-An optional second argument can be passed which will determine the amount of times a split occurs.
+An optional second argument can be passed which will determine the amount of times a split occurs, if negative it's the same
+as not passing a value and will split by all occurrences of the delimiter.
 ```cs
 "Dictu is awesome!".split(" "); // ['Dictu', 'is', 'awesome!']
+"Dictu is awesome!".split(" ", -1); // ['Dictu', 'is', 'awesome!']
 "Dictu is awesome!".split(""); // ["D", "i", "c", "t", "u", " ", "i", "s", " ", "a", "w", "e", "s", "o", "m", "e", "!"]
 "Dictu is awesome!".split(" ", 0); // ['Dictu is awesome!']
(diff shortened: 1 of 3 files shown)


briandowns:
I'm curious about your thoughts on this:

Keep basic functionality and expectations the same.

```cs
"Dictu is awesome!".split(" "); // ['Dictu', 'is', 'awesome!']
```

Add another method, like Go, to indicate the number of times to split.

```cs
"Dictu is awesome!".splitN(" ", 1); // ['Dictu', 'is awesome!']
"Dictu is awesome!".splitN(" ", 2); // ['Dictu', 'is', 'awesome!']
```

or

Have value indicating all instances of the seperator.

```cs
"Dictu is awesome!".split(" ", -1); // ['Dictu', 'is', 'awesome!']
```

Jason2605:
Hey! Thanks for the comments!

> Add another method, like Go, to indicate the number of times to split.

I think another method will probably end up with a lot of code duplication or macro abuse which are never fun to deal with (neither are strings in C at the best of times ;( )

`.split(delim, occur)` is also how it is in Python 😄 

> Have value indicating all instances of the separator.

This is something I don't mind too much about. Checking in Python a negative number seems to make it ignore the count and just split by all occurrences of the delimiter (like how you propose with -1). The current implementation treats negative numbers as not wanting to split (so essentially the same as passing 0). Probably a good idea as it means you can have a negative number as a default rather than having to have two flows in user code with:

```
// Flow 1
var count = cond ? 3 : -1;
".........".split(",", count);

// Flow 2
if (cond) {
    ".........".split(",");
} else {
    ".........".split(",", 3);
}
```

Jason2605:
@briandowns I've updated the PR to make negative numbers split all occurrences now, thanks for the thought!

briandowns:
LGTM!

Issue #490, opened 2021-12-26
[FEATURE] use 'not' instead of '!' for logical negation


gvwilson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Dictu uses `and` and `or` for logical operators; using `not` instead of `!` is more consistent, so remove support for `!` as a synonym for `not` (which has already been added).
### Describe the solution you'd like

1. Minor change to the scanner and compiler (I can do this).
2. Equally minor changes to existing Dictu library, tests, and examples (I can do this).

Jason2605:
Yeah I toyed with this a while ago. Having both obviously isn't great, but at the time my thinking was `!` for "not" made sense as it is similar to `!=`. Having a `not` keyword along with `!=` always was a bit odd for me when working with Python.

Happy to discuss though!

gvwilson:
I guess it's a question of which similarity is more intuitive for new users: `!` as in `!=` or `not` as in `and` and `or`. (I really wish `≠`, `≤`, and `≥` were standard keyboard symbols and part of 7-bit ASCII, but that ship sailed about fifty years ago…)

Jason2605:
Yeah, it's an interesting dilemma and one I'm not sure I know the answer to yet. It also depends what you define "new user" as, if it's someone completely green to programming it's probably less of an issue, but if someone has experience in pretty much any c-like language you're using `!` for `not`.

Appreciate the PR regardless though!

Jason2605:
Since `not` is actually the operator listed in the docs and not `!` I'll probably go ahead with this! Quite bad I hadn't noticed there wasn't both options in the docs previously 😰

PR #492, opened 2021-12-27
Only allowing the word 'not' for logical negation

gvwilson:

- Removed `TOKEN_BANG` from scanner & compiler.
- Removed parsing of `!` as `TOKEN_BANG` so that only the word `not` is accepted as logical negation.
- Updated all affected tests.

Note: some tests are currently failing, but were failing on my system (MacOS 11.6 on Apple M1) before the change as well.

Resolves: #490.
 
### Type of change:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Housekeeping


Jason2605:
> Note: some tests are currently failing, but were failing on my system (MacOS 11.6 on Apple M1) before the change as well.

I've updated the tests so that they should be passing now, I've not attempted a Dictu build on the new M1 chip (i'm still on the intel macbooks), how have you found Dictu on there? Which tests are failing?

Jason2605:
Thanks for this!

PR #495, opened 2021-12-27
initial implementation of chown, chdir, chmod

briandowns:

Initial implementation of chown, chdir, chmod. Some things are easier to test than others. I've updated docs and tests. I need to add additional functions to the System module to ease in testing.

Resolves: 
 
#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping


docs/docs/standard-lib/system.md
```diff
@@ -228,3 +228,35 @@ Shell
...
 $ echo $?; // 10
 ```
+
+### System.chdir(string)
+
+Changes the current working directory to the given path.
+
+```cs
+System.chdir("/usr/local/share");
+```
```
Jason2605, line 238: It's in there twice

```suggestion
```

Jason2605:
Interesting that this PR is causing the Path tests to fail, will need to have a look into this before merging. I'll take a look soon!

Jason2605:
Okay tests are passing on non-windows systems, was due to the fact that chdir wasn't being reset so some of the path tests were failing due to being in the wrong directory.

We will need to cater for windows with these new methods. Either by adding in their equivalent or by noting on the docs that it's not available (I need to double check which are failing).

briandowns:
Looks like `chown` is the culprit here. I'm not overly familiar with Windows to know the best path forward on this one.

Jason2605:
> Looks like `chown` is the culprit here. I'm not overly familiar with Windows to know the best path forward on this one.

Thats fair enough, there are a couple of *nix-only functions already, I'm happy to just hide it behind the `ifdef _win32` guard and just stick a note in the docs saying it's not available. I don't mind doing this if it's easier for you!

briandowns:
> > Looks like `chown` is the culprit here. I'm not overly familiar with Windows to know the best path forward on this one.
> 
> Thats fair enough, there are a couple of *nix-only functions already, I'm happy to just hide it behind the `ifdef _win32` guard and just stick a note in the docs saying it's not available. I don't mind doing this if it's easier for you!

I hate to give you additional work but I think I'll agree so I can get an idea as to how you prefer that to be handled in case I run into that again. Thank you!

Jason2605:
> > > Looks like `chown` is the culprit here. I'm not overly familiar with Windows to know the best path forward on this one.
> > 
> > 
> > Thats fair enough, there are a couple of *nix-only functions already, I'm happy to just hide it behind the `ifdef _win32` guard and just stick a note in the docs saying it's not available. I don't mind doing this if it's easier for you!
> 
> I hate to give you additional work but I think I'll agree so I can get an idea as to how you prefer that to be handled in case I run into that again. Thank you!

Don't worry about it! I'm already grateful for the current input you've done so far!

src/optionals/system.c
```diff
@@ -281,6 +281,77 @@ static Value exitNative(DictuVM *vm, int argCount, Value *args) {
...
+        return EMPTY_VAL;
+    }
+
+    ObjString *file = AS_STRING(args[0]);
+    ObjString *mode = AS_STRING(args[1]);
+
+    int i = strtol(mode->chars, 0, 8);
+
+    if (chmod(file->chars, i) == -1) {
+        char errMsg[1024];
```
Jason2605, line 323: For these parts, can the error length be greater than 256? Do we need to increase this? Just curious why we can't use the `ERROR_RESULT` macro here!

Jason2605:
@briandowns I've updated the PR a bit:
- I've removed `chdir` as we have `setCWD`. I'm unsure about the name of `setCWD` since the rest of the methods in the `System` module don't follow camelCase convention, however, the rest of Dictu does. Maybe need to have a look at this.
- I've made it so `chown` isn't available on windows systems
- I've made it so `chmod` makes use of the proper `_chmod` identifier on windows
- Added / changed a few bits of error handling
- Made it so `chown` returns `nil` rather than `empty` (empty should never be exposed to the user)
- Some house keeping on `setCWD`

Let me know what you think!

Jason2605:
> Looks outstanding. Thank you for the attention on this one.

Awesome, thank you! And thanks for the initial PR!

PR #496, opened 2021-12-27
Env.readFile()

Jason2605:

Relies on: #489 

This PR does two things (I know, I should have really split it up.)
- It fixes a bug where importing something from the stdlib that was formed with Dictu code would segfault (UnitTest library for example)
- Adds `Env.readFile()` which is a helper method to read and parse env files. By default it'll read `.env` unless supplied another path. It takes into account `#` comments for both starting the line and in-line.

Resolves: #473 

### Type of change:

- [x] Bug fix

- [x] New feature

#

### Housekeeping


Jason2605:
I'd be interested on your thoughts on the `readFile` implementation @briandowns 

See here: https://github.com/dictu-lang/Dictu/blob/90f6f12f0425e77dba7c6513861c22470984a29e/src/optionals/env/env.du

briandowns:
This looks really clean. I dig it!

PR #500, opened 2021-12-29
Add openweathermap example app

briandowns:

Adds example code using the HTTP, System, Env, and JSON modules to connect to, parse, and print a response from openweathermap.org.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

#

### Housekeeping

#

### DIFF
--- a/examples/openweathermap.du
+++ b/examples/openweathermap.du
@@ -1,0 +1,49 @@
+/**
+ * Example of how to use the HTTP module and JSON modules to interact with the 
+ * OpenWeatherMap REST API. This can serve as a base to extend with more features.
+ */
+
+import Env;
+import HTTP;
+import JSON;
+import System;
+
+const BASE_URL = "http://api.openweathermap.org/data/2.5/weather?";
+
+class Client {
+    private token;
+    
+    init(units, lang, token) {
+        this.units = units;
+        this.lang = lang;
+        this.token = token;
+    }
+
+    current(name) {
+        var url = BASE_URL;
+        url = url + "appid=" + this.token;
+        url = url + "&q=" + name;
+        url = url + "&units=" + this.units;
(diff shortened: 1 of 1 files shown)


examples/openweathermap.du
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    current(name) {
+        var url = BASE_URL;
+        url = url + "appid=" + this.token;
+        url = url + "&q=" + name;
+        url = url + "&units=" + this.units;
+        url = url + "&lang=" + this.lang;
+
+        var res = HTTP.get(url);
```
Jason2605, line 29:
```suggestion
        const res = HTTP.get(url);
```

examples/openweathermap.du
```diff
@@ -0,0 +1,49 @@
...
+        return JSON.parse(res.unwrap()["content"]).unwrap();
+    }
+}
+
+// main
+{
+    const token = Env.get("OWM_TOKEN");
+    if (token == "" or not token) {
```
Jason2605, line 37: An empty string is considered "falsey" so all we need is the `not` call!

```suggestion
    if (not token) {
```

Jason2605 (changes requested):
More incredibly nitpick comments sorry!

Also if we add examples to the `runExamples.du` file, it'll make sure they don't become outdated: https://github.com/dictu-lang/Dictu/blob/develop/examples/runExamples.du

briandowns:
This example is dependent on a valid token being available set as an env var. Not sure if we'd want to add it to `runExamples.du`.

Jason2605:
> This example is dependent on a valid token being available set as an env var. Not sure if we'd want to add it to `runExamples.du`.

Fair point yep!

PR #504, opened 2021-12-29
add template engine example app

briandowns:

Adds a very naive template engine to show off use of `getAttributes`.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

#

### Housekeeping

#

### DIFF
--- a/examples/template_engine.du
+++ b/examples/template_engine.du
@@ -1,0 +1,60 @@
+import System;
+
+const LEFT_BRACE = "{";
+const RIGHT_BRACE = "}";
+
+// Template is a simple template rendering engine.
+class Template {
+    private tmpl;
+    private klass;
+    private rendered;
+
+    init(tmpl, klass) {
+        this.tmpl = tmpl;
+        this.klass = klass;
+    }
+
+    // render parses the given template and class and
+    // matches the fields in the class to the template
+    // fields and replaces them with the class values.
+    render() {
+        const classAttrs = this.klass.getAttributes()["properties"];
+        this.rendered = this.tmpl;
+
+        classAttrs.forEach(def(attr) => {
+            const attrVal = this.klass.getAttribute(attr);
+            const tmplField = LEFT_BRACE+attr+RIGHT_BRACE;
+
+            if (this.rendered.contains(tmplField)) {
+                if (type(attrVal) != "string") {
+                    this.rendered = this.rendered.replace(tmplField, attrVal.toString());
+                } else {
+                    this.rendered = this.rendered.replace(tmplField, attrVal);
+                }
+            }
+        });
+
+        return this.rendered;
+    }
+}
+
+// Person is a class used to hold data to be passed 
+// to the template engine.
+class Person {
+    init(name, age) {
+        this.name = name;
+        this.age = age;
+    }
+}
+
+// main
+{
+    const tmpl = "Hello {name}! You are {age} years old.";
+    const p = Person("John", 12);
+
+    const t = Template(tmpl, p);
+
+    print("{}".format(t.render()));
+
+    System.exit(0);
+}
--- a/examples/template_engine.du
+++ b/examples/template_engine.du
@@ -5,10 +5,7 @@ class Template {
     var LEFT_BRACE = "{";
     var RIGHT_BRACE = "}";
 
-    init(tmpl, klass) {
-        this.tmpl = tmpl;
-        this.klass = klass;
-    }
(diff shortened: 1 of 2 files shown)


examples/template_engine.du
```diff
@@ -0,0 +1,58 @@
...
+// Template is a simple template rendering engine.
+class Template {
+    private tmpl;
+    private klass;
+    private rendered;
+
+    init(tmpl, klass) {
+        this.tmpl = tmpl;
+        this.klass = klass;
+    }
```
Jason2605, line 15: Just some suggestions here, open for discussion.

- I've moved the `LEFT_BRACE` / `RIGHT_BRACE` constants inside the class to be class variables rather than constants outside of the class
- I've removed the `rendered` private property. My rationale for this is it's only ever used inside render and is defined as private. If it was public it could potentially have use for caching, but as it's private it doesn't need to be on the object.
- Ive removed the explicit private / initialisation for tmpl and klass for their implicit property counterparts.

```suggestion
// Template is a simple template rendering engine.
class Template {
    var LEFT_BRACE = "{";
    var RIGHT_BRACE = "}";

    init(private tmpl, private klass) {}
```

examples/template_engine.du
```diff
@@ -0,0 +1,58 @@
...
+                    this.rendered = this.rendered.replace(tmplField, attrVal.toString());
+                } else {
+                    this.rendered = this.rendered.replace(tmplField, attrVal);
+                }
+            }
+        });
+
+        return this.rendered;
+    }
+}
```
Jason2605, line 39: Here I've done:
- Changed `this.rendered` to be variable `rendered` local to `render()`
- Changed the formatting for `tmplField` to use `.format` with the new class variables

```suggestion
    render() {
        const classAttrs = this.klass.getAttributes()["properties"];
        var rendered = this.tmpl;

        classAttrs.forEach(def(attr) => {
            const attrVal = this.klass.getAttribute(attr);
            const tmplField = "{}{}{}".format(
                Template.LEFT_BRACE,
                attr,
                Template.RIGHT_BRACE
            );

            if (rendered.contains(tmplField)) {
                if (type(attrVal) != "string") {
                    rendered = rendered.replace(tmplField, attrVal.toString());
                } else {
                    rendered = rendered.replace(tmplField, attrVal);
                }
            }
        });

        return rendered;
    }
```

examples/template_engine.du
```diff
@@ -0,0 +1,58 @@
...
+
+// main
+{
+    const tmpl = "Hello {name}! You are {age} years old.";
+    const p = Person("John", 12);
+
+    const t = Template(tmpl, p);
+
+    print("{}".format(t.render()));
```
Jason2605, line 57: Here we can just print directly, no string formatting required

```suggestion
    print(t.render());
```

Jason2605 (commented):
I've left a few comments for discussion, let me know what you think!

examples/template_engine.du
briandowns: All of that seems fine. Makes me wonder how hard it would be to implement constants on classes. Or even a const key word in function args.

Jason2605: I think constants on classes shouldn't be _too_ bad. It would pretty much be the same flow as class vars currently but potentially to a new internal Table within the class.

e.g: https://github.com/dictu-lang/Dictu/blob/develop/src/vm/object.h#L223

We could add
```c
Table constantProperties;
```

https://github.com/dictu-lang/Dictu/blob/develop/src/vm/vm.c#L1254-L1260

Here we could extend the opcode where `var` would add to `publicProperties` and `const` would add to `constantProperties`.

https://github.com/dictu-lang/Dictu/blob/develop/src/vm/vm.c#L1014-L1022

The following bit of code would need to be extended to also check for constants (in all places it occurs)

```c
while (klass != NULL) {
                        if (tableGet(&klass->publicProperties, name, &value)) {
                            pop(vm); // Instance.
                            push(vm, value);
                            DISPATCH();
                        }

                        if (tableGet(&klass->constantProperties, name, &value)) {
                            pop(vm); // Instance.
                            push(vm, value);
                            DISPATCH();
                        }

                        klass = klass->superclass;
                    }
```

Finally here: https://github.com/dictu-lang/Dictu/blob/develop/src/vm/vm.c#L1229-L1236

We would probably need to check if the property attempting to be assigned to the class isn't already present in the constants table

Jason2605:
Thanks for the PR!

Issue #508, opened 2021-12-31
[FEATURE] - Allow for the use of constants in classes


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Currently, the only way to use a `const` is out of the context of a class. It'd be nice to have the same functionality and guarantees in classes as well.

PR #510, opened 2022-01-17
Class constants

Jason2605:

This PR adds the ability to define a constant within a class. This comes with the runtime guarantee that the value will not be changed.

E.g
```
class SomeClass {
    const classVariable = 10; // This will be shared among all "SomeClass" instances

    init() {
        this.x = 10; // "x" is set on the instance
    }
}

print(SomeClass.classVariable); // 10

const x = SomeClass();
const y = SomeClass();

print(x.classVariable); // 10
print(y.classVariable); // 10

SomeClass.classVaraible = 100; // Cannot assign to class constant 'SomeClass.classVariable'.
```

Resolves: #

#508 

### Type of change:

- [x] New feature

#

### Housekeeping

#

### DIFF
--- a/docs/docs/classes.md
+++ b/docs/docs/classes.md
@@ -367,7 +367,7 @@ const y = SomeClass();
 print(x.classVariable); // 10
 print(y.classVariable); // 10
 
-SomeClass.classVaraible = 100; // Cannot assign to class constant 'SomeClass.classVariable'.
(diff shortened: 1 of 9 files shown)


briandowns:
This looks awesome!

Jason2605:
Thanks @briandowns!

I was on the fence about this due to the fact that it only provides runtime guarantees (const in a variable(I know that doesn't quite make sense, but hopefully you get what I mean) are detected at compile time).

Probably still worthwhile however

Issue #341, opened 2020-12-01
Shadowing builtins has a weird effect


Jason2605:
# Shadowing
## Summary
```
var input = [];
print(input); // <native fn>
```

Here we would expect the variable to take precedence over the built-in function, although shadowing is bad either way, the principle of least surprise would be that the built-in is overwritten, however currently it is not.

PR #523, opened 2022-05-03
Fix a niche issue with global shadowing

Jason2605:

This change makes it so that shadowing a builtin in the module level space has the same effect as shadowing a builtin in the local level space. Previously setting a variable with an identifier of a builtin (e.g input) in the module level space would keep the value of the builtin value rather than taking the new value of the variable. Shadowing should be avoided, but this at least keeps the scoping consistent.

Resolves: #341 

#

### Type of change:

- [x] Bug fix

#

### Housekeeping

#

### DIFF
--- a/src/vm/compiler.c
+++ b/src/vm/compiler.c
@@ -1191,8 +1191,9 @@ static void namedVariable(Compiler *compiler, Token name, bool canAssign) {
     } else {
         arg = identifierConstant(compiler, &name);
         ObjString *string = copyString(compiler->parser->vm, name.start, name.length);
-        Value value;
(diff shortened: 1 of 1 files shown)


Jason2605:
This will need revising, tested via the REPL and works, however, executing a file it will not work. This is due to the fact that the variable would not have been added to the VM module until runtime and we are checking at compile time. Works in the REPL as each line is a new compilation

RevengerWizard:
Hi (again), I have a similar issue also in my little language. The problem isn't really in the finding of the variable, but more of how var declaration works, since it always leads to a module definition (or local), so it creates the variable at module scope, but it will always be shadowed by the global definition. Or at least, that's what I was thinking.

PR #534, opened 2022-09-29
Adds repeat Method to string Type

briandowns:

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

#


src/vm/datatypes/strings.c
```diff
@@ -530,6 +530,29 @@ static Value titleString(DictuVM *vm, int argCount, Value *args) {
...
+    int count = AS_NUMBER(args[1]);
+    char *temp = ALLOCATE(vm, char, (string->length * count) + 1);
+
+    strcpy(temp, string->chars);
+    while (--count > 0) {
+        strcat(temp, string->chars);
+    }
+
+    return OBJ_VAL(takeString(vm, temp, strlen(temp)));
+}
```
Jason2605, line 554: Codacy is complaining about the use of `strlen` and that it could potentially cause issues if the string it's counting is not properly null terminated. In our case though, we know for definite the length of our string, so we can re-use this count in the `ALLOCATE` macro and the `takeString` function rather than re-calculating!

```suggestion
    int length = (string->length * count) + 1;
    char *temp = ALLOCATE(vm, char, length);

    strcpy(temp, string->chars);
    while (--count > 0) {
        strcat(temp, string->chars);
    }

    return OBJ_VAL(takeString(vm, temp, length - 1));
}
```

tests/strings/repeat.du
```diff
@@ -0,0 +1,17 @@
+/**
```
Jason2605, line 1: We will need to make sure we import this new file to the tests/strings/import.du file, there's currently no smart file resolution for the tests

https://github.com/dictu-lang/Dictu/blob/develop/tests/strings/import.du

Jason2605 (changes requested):
Thank you very much for the PR @briandowns! Just a couple of comments

briandowns:
All set now. Thanks!

Jason2605:
Looks great, thank you!

PR #536, opened 2022-10-01
add mkdirTemp to System module

briandowns:

Adds the ability to create temporary directories.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

#


src/optionals/system.c
```diff
@@ -100,6 +100,36 @@ static Value chownNative(DictuVM *vm, int argCount, Value *args) {
...
+
+    if (template[0] != '\0') {
+        tmpl = malloc(strlen(template)+1);
+        strcpy(tmpl, template);
+    } else {
+        tmpl = malloc(PATH_MAX);
+        strcpy(tmpl, "XXXXXX");
+    }
+
+    char *tmpDir = mkdtemp(tmpl);
```
Jason2605, line 126: I think it's worth using the Result type here as mkdtemp can fail. In those situations it would be good to give some indication as to why it failed, we can do that by unwrapping a Result type

src/optionals/system.c
```diff
@@ -100,6 +100,36 @@ static Value chownNative(DictuVM *vm, int argCount, Value *args) {
...
+    if (!IS_STRING(args[0])) {
+        runtimeError(vm, "mkdirTemp() first argument must be a string");
+        return EMPTY_VAL;
+    }
+
+    char *template = AS_CSTRING(args[0]);
+    char *tmpl = {0};
+
+    if (template[0] != '\0') {
+        tmpl = malloc(strlen(template)+1);
```
Jason2605, line 119: I try to avoid uses of malloc / alloc etc where possible and try to use the memory macros, this means if in the future we ever needed to change the memory allocator it's a simple switch.

We'll need to ensure we're freeing things up too since on success where creating a new string rather than inheriting the already alloc'd one.

Jason2605 (changes requested):

src/optionals/system.c
```diff
@@ -162,6 +162,36 @@ static Value unameNative(DictuVM *vm, int argCount, Value *args) {
...
+    char *template = AS_CSTRING(args[0]);
+    char *tmpl = {0};
+
+    if (template[0] != '\0') {
+        tmpl = ALLOCATE(vm, char, strlen(template));
+        strcpy(tmpl, template);
+    } else {
+        tmpl = ALLOCATE(vm, char, PATH_MAX);
+        strcpy(tmpl, "XXXXXX");
+    }
```
Jason2605, line 186: We need to allocate for the terminating char too. My thinking here also is that if no template is passed the directory would be made via the 6 digits of XXXXXX template right? Doesn't that mean we only need to allocate for these 6 + \0?

```suggestion
    if (template[0] != '\0') {
        tmpl = ALLOCATE(vm, char, strlen(template) + 1);
        strcpy(tmpl, template);
    } else {
        tmpl = ALLOCATE(vm, char, 7);
        strcpy(tmpl, "XXXXXX");
    }
```

Jason2605: We could also probably keep reference to the length rather than calculating again via strlen in the takeString call

Jason2605:
I see it's failing on the ubuntu machines, I'll need to give that a test and see whats going on but I don't have access to a machine until later on in the week now sorry!

Jason2605:
Nevermind, found the issue! The test didn't setup the template to end with `XXXXXX` so it was retuning an error result, at this point the memory wasn't being free'd! I've also changed a couple things (rather than an empty string allow no template to be passed for ease) and added a few notes to the docs page. Think we should be all good on this PR now, thank you!

briandowns:
Thanks. Been swamped this week. Was gonna get back to it this weekend.

Jason2605:
No worries at all! I'd just noticed the issue after another quick scan of the PR and thought I'd get it in. Appreciate the PRs you've been opening!

PR #538, opened 2022-10-09
Added containsAll() method for sets

MannarAmuthan:
containsAll() method for set.

### set.containsAll(value)

To check if a set contains all elements in a given list use `.containsAll()`

```cs
var mySet = set("one",1,2,3);;
print(mySet.containsAll(["one",1])); // true
print(mySet.containsAll([1,2,3])); // true
print(mySet.containsAll(["one",1,2,3,"x"])); // false
```

- [ Done] Added unit tests
-  [Done ] Updated Doc


Jason2605:
Thanks for this, just one comment, would it be possible to rename it to containsAll please? I think that better reflects what this does and removes ambiguity about it returning a truthy value if it contains **some** values

MannarAmuthan:
Sure will do that 👍

Jason2605:
Thanks for this!

PR #539, opened 2022-10-10
Object.createFrom()

Jason2605:

This PR adds a new module `Object` that allows us to retrieve class references via strings or objects to be created via strings rather than always having a reference to the class.

### Type of change:

- [x] New feature

#

### Housekeeping


docs/docs/standard-lib/object.md
```diff
@@ -0,0 +1,53 @@
...
+This method will attempt to get a class reference from the class name provided as a string.
+
+Returns a Result and unwraps to an Object upon success.
+
+**NOTE:** This currently only works if the class is defined in the global scope
+
+```cs
+class Test {}
+
+Object.getClassRef("Test").unwrap(); // <Cls Test>
```
briandowns, line 38: Is there an easy way to get the name of the class from an object? I know we have a getAttributes method but I don't believe it includes the name of the class. That might be a nice addition so this can be even more dynamic.

Jason2605: When you say class name, do you mean an actual reference to the class or do you mean the string representation of the class name?

briandowns: The string representation of the class.

Jason2605: Ah there's currently no way to turn a class reference into a string'd name. We could add a name `_name` attribute similar to `_class` to classes

briandowns: That would be cool.

Jason2605: Should be added as part of #543 

Now it would be:
```
class Test {}

print(Test._name);
print(Test()._class._name);
```

briandowns: Very cool!

Issue #421, opened 2021-09-30
List unpacking


Jason2605:
# List unpacking
## Summary
List unpacking is only working when we are declaring the variables. It should also work with previously defined variables. Need to investigate further into why this is happening, will drop more info here as I find it.

#### Example
![image](https://user-images.githubusercontent.com/20248039/135511756-e4697d1e-c90d-4ec8-a2a3-4155ad811737.png)

Fly-Style:
I'd like to pick it.

Jason2605:
@Fly-Style awesome, assigned this to you!

Jason2605:
@Fly-Style Going to remove your assignment as it's been just over half a year now.

MannarAmuthan:
Hi @Jason2605,   Could you provide the example for a proper **list unpacking**?

Jason2605:
@MannarAmuthan There is an example in the screenshot in how it should work, but doesn't, it currently works when you declare a variable but not when you set a variable.

Example:

```
// Fine
var [x, y] = [1, 2];
print(x); // 1
print(y); // 2

// Issues
var x, y;
[x, y] = [1, 2]; // Invalid assignment target
```

MannarAmuthan:
Hi Could you assign this to me?, I'd like to pick it.

MannarAmuthan:
Hi , Which one is correct syntax for unpacking already declared variables x,y?

var x,y;
[x, y] = [1, 2]; 
(or)
var x,y;
x,y=[1,2];

Jason2605:
The first option is what we should have available to us for list unpacking

```
var x,y;
[x, y] = [1, 2];
```

MannarAmuthan:
Raised pull request for fix: https://github.com/dictu-lang/Dictu/pull/546

MannarAmuthan:
Found some issues , will fix it

PR #546, opened 2022-10-20
Fixed list unpacking issue

MannarAmuthan:
Fix for issue: https://github.com/dictu-lang/Dictu/issues/421 

Found that we need to add separate statement parsing logic for already declared variables to list unpacking .

[X] -  Added this scenario in tests


MannarAmuthan:
Found some issue, fixing it

MannarAmuthan:
I still get the error on this example tests, even when I revert my changes

Jason2605:
Yeah it's due to the fact that we can't use list literals. This issue is actually going to be rather tricky to solve I think

Screenshot 2022-10-20 at 17 36 26

tests/variables/list-unpacking.du
```diff
@@ -9,6 +9,20 @@ class TestListUnpacking < UnitTest {
...
+        var o,i,g;
+
+        [o,i,g]=[111,222,333];
+
+        assert(o == 111);
+        assert(i == 222);
+        assert(g == 333);
+
+        var [d, e, f] = [1 + 1, 2 + 2, 3 + 3];
+
```
Jason2605, line 25: We want to use the https://dictu-lang.com/docs/standard-lib/unittest/ builtin where possible over `assert` as `assert` will stop execution on first failure (as the main reason - UnitTest library has a nicer output too).

MannarAmuthan:
This is Done

Jason2605:
Awesome bit of work @MannarAmuthan, thanks very much for this!

PR #547, opened 2022-10-22
Fixed self import Segmentation fault

manoharkakumani:

   While Importing the same file as a module in the same file we are getting a Segmentation fault.
Ex: if we execute the below code by writing it in test.du file
`import "./test.du";
var x = 1;
print(x);`

Resolves: #

  After creating the module we need to add it to vm->modules table to keep it tracked.
#

### Type of change:

 added below line in vm.c file in IMPORT opcode block
`tableSet(vm, &vm->modules, fileName, OBJ_VAL(module));`

- [ ] Bug fix

- [ ] New feature

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

#

### Preview (Screenshots) :

Before changes

image

After Change

image

### DIFF
--- a/src/vm/vm.c
+++ b/src/vm/vm.c
@@ -1589,7 +1589,7 @@ static DictuInterpretResult run(DictuVM *vm) {
             module->path = dirname(vm, path, strlen(path));
             vm->lastModule = module;
             pop(vm);
-
+            tableSet(vm, &vm->modules, fileName, OBJ_VAL(module));
             push(vm, OBJ_VAL(module));
             ObjFunction *function = compile(vm, module, source);
             pop(vm);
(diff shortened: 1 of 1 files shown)


Jason2605:
This is a good spot, thank you! The fix however, we will want to be slightly different. Rather than saving the filename, we should instead check for the path as this is what is being saved in newModule.

manoharkakumani:
Have you fixed the issue? If yes I'll close the pr.
Can you please let me know where you are setting the vm->modules.

Jason2605:
I haven't committed anything, but the above code would resolve the issue yeah. The module gets added in the call to `newModule`, the issue was we were checking for the filename given by the user, not the fully resolved file name! So good spot on your part

See here: https://github.com/dictu-lang/Dictu/blob/develop/src/vm/object.c#L45

Jason2605:
I've now pushed the above changes to your PR, and I'll merge it in now. This means if you're participating in hacktoberfest or similar you can still get the credit 🎉

PR #550, opened 2022-11-05
Term module implementation

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

### Well detailed description of the change :

Adds a new module "Term" with functions to get information about the attached terminal.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

- [ ] Tests have been updated to reflect the changes done within this PR (if applicable).

- [ ] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

PR #549, opened 2022-11-02
Fixed self referencing datatypes bug

manoharkakumani:
### Well detailed description of the change :

updated listToString and dictToString functions inorder to skip segfault

```
// for list
var x = [1,2,3];
x.push(x);
print(x);// [1,2,3,[...]
```
```
// for dict
var a ={1:2};
a[2]=a;
print(a);// {2: {...}, 1: 2}
```

Resolves: #
  Self referencing datatypes #354
#

### Type of change:

- [x] Bug fix

- [ ] New feature

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

- [ ] Tests have been updated to reflect the changes done within this PR (if applicable).

- [ ] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

### Preview (Screenshots) :

image

PR #548, opened 2022-10-31
HTTP Client Implementation

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

### Well detailed description of the change :

Adds an HTTP client that is configurable and reusable.

Resolves: #

#537 

### Type of change:

- [ ] Bug fix

- [x] New feature

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

- [ ] Tests have been updated to reflect the changes done within this PR (if applicable).

- [ ] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

PR #545, opened 2022-10-20
add proxy example

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

### Well detailed description of the change :

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

- [ ] Tests have been updated to reflect the changes done within this PR (if applicable).

- [ ] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

PR #553, opened 2022-11-07
Release 0.26.0

Jason2605:
# Release 0.26.0
## Change Log
**These are all the changes since 0.25.0:**
- Update README to include FreeBSD build steps #527 
- Add argument unpacking at callsites (spread operator) #528 
- Resolve an issue where expressions in the REPL were being printed when it wasn't intended #530 
- Add `string.repeat(num)` #534 
- Add `uname` to System module #535 
- Add `mkdirTemp` to System module #536 
- Add `set.containsAll()` #538 
- Add a new builtin Module `Object` #539 
    - Object.createFrom
    - Object.getClassRef
- Fixed a bug where `break` used within a function defined within a loop would cause a segfault #542 
- Add `_name` attribute to classes which returns the class name as a string #543 
- Add `Env.clearAll()` #544 
- Add Dictu proxy server example #545 
- Fixed an issue where you could not unpack lists unless declaring the variable #421 
- Fixed an issue with imports saving the pathname incorrectly #547 
- Add new client to HTTP module #548 
- Fix an issue where self referencing datatypes would cause a segfault when attempting to print #549 
- Add new `Term` module #550


briandowns:
Not sure your thoughts but I only have this last thing in the term module. https://github.com/dictu-lang/Dictu/pull/550#discussion_r1014673637

I can knock that out and we can include that module in the new release.  No big deal either way. Just figured I'd mention it.

Jason2605:
Looks like we're good to go on this, massively appreciate the work @briandowns and @MannarAmuthan!

briandowns (approved):
LGTM!

Issue #564, opened 2022-11-15
[FEATURE] - HTTP Client Always Follows Redirects


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Most HTTP clients in languages will follow redirects however provide a means of telling the client not to do so.

### Describe the solution you'd like

Add a field to the http client options indicating that the client it is to not follow redirects.

Jason2605:
In my testing it seems it's currently defaulting to the opposite (as in not following redirects). I wonder if this is something we should default to true throughout the library and allow people to opt-out with the client options

briandowns:
Yes, sorry. I didn't word that very well. Yes, that's the current behavior. I'll submit a PR this evening.

PR #566, opened 2022-11-16
add follow redirects to http

briandowns:

Adds default behavior to follow redirects and adds ability to turn that behavior off on the client.

Resolves: #564 

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

#

### Housekeeping

#

### DIFF
--- a/src/optionals/http/http.c
+++ b/src/optionals/http/http.c
@@ -466,6 +466,7 @@ static Value get(DictuVM *vm, int argCount, Value *args) {
         curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response);
         curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, writeHeaders);
         curl_easy_setopt(curl, CURLOPT_HEADERDATA, &response);
+        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
 
         /* Perform the request, res will get the return code */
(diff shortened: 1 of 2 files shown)


src/optionals/http/http.c
```diff
@@ -948,6 +973,19 @@ ObjAbstract *newHttpClient(DictuVM *vm, ObjDict *opts) {
...
                     curl_easy_setopt(httpClient->curl, CURLOPT_SSL_VERIFYPEER, 0);
                 }
+            } else if (strstr(key, "follow_redirects")) {
+                if (IS_EMPTY(entry->value)) {
+                    continue;
+                }
+
+                if (!IS_BOOL(entry->value)) {
+                    runtimeError(vm, "HTTP client option \"follow_redirects\" value must be a bool");
+                    return abstract;
```
Jason2605, line 983: I've just realised an issue here. We aren't handling the failure as we should here, this will probably end up with a segfault in the erroneous scenario.

Instead what we need to do is whenever we raise a runtimeError we return an `EMPTY_VAL`. So `newHttpClient` should look like: `Value newHttpClient(DictuVM *vm, ObjDict *opts)`

And it means `newClient` can just return as:
```
return newHttpClient(vm, opts);
```

briandowns: All set I believe.

Jason2605:
Thanks for this!

PR #577, opened 2022-12-07
Add UUID Module

briandowns:

Added a UUID module that basically wraps the common C UUID library.

Not really sure where to put the uuid library dependency. Also, I might need to make this *NIX only.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

#


Jason2605:
It'll be failing on ubuntu as libuuid hasn't been linked to it in the CMake file, I think it'll be good to have this behind a flag too incase we ever want to compile without linking libuuid

An example with cURL:
https://github.com/dictu-lang/Dictu/blob/develop/src/CMakeLists.txt#L15-L21

Jason2605:
Sorry, super slow on these recently! I assume none of the methods in the UUID library are available on windows systems?

briandowns:
No. Sadly.

Jason2605:
Thanks for this!

PR #587, opened 2023-01-11
Use nfpm to Create RPM, deb, apk Packages, and Upload Linux Binary

briandowns:

Add the ability to create RPMs, debs, and apks as well as upload the compiled binary for easy download and use per release.

Depending on if this is something we want to do, we'd need to update the github action to create the packages and upload them to the release when new releases are created.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

#


Jason2605:
This looks nice thank you! I've not personally used nFPM before but will take a look into it, thanks. Adding files to a release should be fine, there are a couple of actions available for it on the market place so we can use them / draw inspiration from those

briandowns:
I think this is ready to go.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+        ./dictu ops/pkgBuild.du
+        mv dictu dictu-linux-x86_64
+    - name: Release
+      uses: SierraSoftworks/gh-releases@v1.0.7
+      with:
+        token: ${{ secrets.TOKEN }}
+        overwrite: 'true'
+        files: |
+          dictu-linux-x86_64
+          dictu-0.26.0-1_amd64.deb
```
Jason2605, line 33: We will probably need to think of a way to make this version dynamic

ops/nfpm.yaml
```diff
@@ -0,0 +1,122 @@
...
+# This will expand any env var you set in the field, e.g. version: ${GOOS}
+# This is only used by the rpm and deb packagers.
+# Examples: `linux` (default), `darwin`
+platform: linux
+
+# Version. (required)
+# This will expand any env var you set in the field, e.g. version: ${SEMVER}
+# Some package managers, like deb, require the version to start with a digit.
+# Hence, you should not prefix the version with 'v'.
+version: 0.26.0
```
Jason2605, line 22: Here too

Jason2605 (commented):
Cheers for this!

Jason2605:
Thanks for this! Due a release soon so will see what happens

Issue #567, opened 2022-11-16
[FEATURE] - Provide Ability to Write Downloaded Data to a File


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

There doesn't appear to be any way of writing the content retrieved from an HTTP call to a file. I don't have any good suggestions on how to implement this as there are a number of ways to examine. Open to suggestions.

Jason2605:
Are you talking about some sort of stream interface for large files or?

briandowns:
Yes, for instance downloading tars, or other files of interest.

briandowns:
I've been giving this some thought and I haven't come up with anything concrete. I wanted to come up with something that wouldn't break the current API but that seems impossible.

Option 1:
Add an optional argument to argument to the `get` methods of a file path and on existence, a function is given to curl to write the received data to that file.

Option 2:
Introduce a new type to Dictu called `Bytes` and change the HTTP.Response.content field to be of this type. The caller then chooses what to do with the bytes. We could also add `b'all of the bytes';` style 

Option 3:
Add a method to the module and on the client to specifically download files.

Option 4:
Another much better option.

briandowns:
I was hoping to work on this issue next week during my company's hackweek. Wanted to get your (@Jason2605 ) thoughts on how you want to see this implemented. Thanks in advance!

Jason2605:
I think option 2 is probably the route I would go with introducing a new bytes type, as this is something that will need doing eventually anyways. Once we have a dedicated bytes type we can default the string type to handle unicode chars to bring the string type into the modern era. Probably the most work involved in this option, but I think it's the way to go!

briandowns:
Thank you for the feedback! Will get started on this next week.

Jason2605:
Much appreciated, thank you

briandowns:
I found a workout around for this that will not require any new types or adjustments to the HTTP module. I'll submit the PR by tomorrow night. This will open the door to so much. :D

PR #596, opened 2023-02-05
update file handling to account for writing in binary mode

briandowns:

Adds the ability to write in binary mode.

With this change, the following code works:

```cs
#!/usr/bin/env dictu

import HTTP;
import System;

{ // main
    const opts = {
        "timeout": 20,
        "headers": [
            "User-Agent: Dictu"
        ],
        "insecure": true,
        "follow-redirects": true
    };
        
    const client = HTTP.newClient(opts);

    const res = client.get("https://curl.se/download/curl-7.86.0.zip");
    const result = res.unwrap();

    with("curl-7.86.0.zip", "wb") {
        file.write(result.content);
    }

    System.exit(0);
}
```

Resolves: #

#567 

### Type of change:

- [ ] Bug fix

- [x] New feature

#

### Housekeeping

#

### DIFF
--- a/src/vm/datatypes/files.c
+++ b/src/vm/datatypes/files.c
@@ -15,13 +15,19 @@ static Value writeFile(DictuVM *vm, int argCount, Value *args) {
     ObjFile *file = AS_FILE(args[0]);
     ObjString *string = AS_STRING(args[1]);
 
-    if (strcmp(file->openType, "r") == 0) {
+    if (strcmp(file->openType, "r") == 0 || strcmp(file->openType, "rb") == 0) {
         runtimeError(vm, "File is not writable!");
         return EMPTY_VAL;
     }
 
-    int charsWrote = fprintf(file->file, "%s", string->chars);
-    fflush(file->file);
+    int charsWrote = 0;
+
+    if (strcmp(file->openType, "wb") == 0) {
+        charsWrote = fwrite(string->chars, 1, string->length, file->file);
+    } else {
+        charsWrote = fprintf(file->file, "%s", string->chars);
+        fflush(file->file);
+    }
 
     return NUMBER_VAL(charsWrote);
 }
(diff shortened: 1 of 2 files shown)


src/vm/datatypes/files.c
```diff
@@ -15,13 +15,19 @@ static Value writeFile(DictuVM *vm, int argCount, Value *args) {
...
 
-    int charsWrote = fprintf(file->file, "%s", string->chars);
-    fflush(file->file);
+    int charsWrote = 0;
+
+    if (strcmp(file->openType, "wb") == 0) {
+        charsWrote = fwrite(string->chars, 1, string->length, file->file);
+    } else {
+        charsWrote = fprintf(file->file, "%s", string->chars);
+        fflush(file->file);
```
Jason2605, line 29: Probably worth moving the flush call outside the if blocks so it's called regardless

src/vm/datatypes/files.c
```diff
@@ -82,7 +88,10 @@ static Value readFullFile(DictuVM *vm, int argCount, Value *args) {
         buffer = SHRINK_ARRAY(vm, buffer, char, fileSize + 1, bytesRead + 1);
     }
 
-    buffer[bytesRead] = '\0';
+    if (strcmp(file->openType, "rb") != 0) {
+        buffer[bytesRead] = '\0';
+    }
```
Jason2605, line 93: We will want to null terminate here as Dictu expects all strings to be null terminated. Is there a reason you didn't, did you find something break or?

briandowns:
I clicked on the "request review" button only to let you know the changes are in. No rush at all. I saw in another PR you said you have a busy week so figured I'd drop this note so not to seem pushy. :D

Jason2605:
Haha I would prefer to get the extra emails than miss something, so cheers for the reminder! Yeah busy one this week, will try to get round to it

Jason2605:
Thanks for this

PR #598, opened 2023-02-06
Removed option to build without linenoise (depends on #597)

jamsilva:
### Well detailed description of the change :

Removed DISABLE_LINENOISE macro and relevant code.

#

### Type of change:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping

- [x] Documentation has been updated to reflect the changes done within this PR (if applicable).

PR #599, opened 2023-02-07
REPL fixes (depends on #598)

jamsilva:

This corrects multiline statements in the REPL in 2 ways:
1) The combined multiline string is now always null-terminated - for example, copying the contents of fibonacci.du would previously cause an 'Unexpected character' error.
2) The lines have newlines separating them, so the error messages show the line number of the error. The following error would previously show line 1 for everything.
```
>>> {
... {
... 'a'
... }
... }
Error: File 'repl', line 4
       4 | }
       Expect ';' after expression.

Error: File 'repl', line 5
       Error at end: Expect '}' after block.

Error: File 'repl', line 5
       Error at end: Expect '}' after block.
```

The string literal and brace matching code was also changed to act more like the way a parser would do it.

#

### Type of change:

- [x] Bug fix

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -13,53 +13,59 @@
 
 #include "linenoise/linenoise.h"
 
-static bool replCountBraces(char *line) {
-    int leftBraces = 0;
-    int rightBraces = 0;
-    bool inString = false;
-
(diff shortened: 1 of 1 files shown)


Jason2605:
This is really great, thank you very much for this!

Issue #608, opened 2023-03-06
[BUG] Upvalues aren't closed correctly after `continue` and `break` statements


RevengerWizard:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, it seems that `continue` and `break` statements are broken when inside of a loop body there is a closure. This is because currently both statements when closing all locals don't check if the current local popped is an upvalue.

```c
// Discard any locals created inside the loop.
for (int i = compiler->localCount - 1;
     i >= 0 && compiler->locals[i].depth > compiler->loop->scopeDepth;
     i--) {
    emitByte(compiler, OP_POP);
}
```

This piece of code is present in both `continueStatement` and `breakStatement` functions.
I'd just open a pull request, but I'm sadly I'm unable to correctly compile Dictu under Windows with MinGW because of some weird compiler error, so I'll just provide the fix here:

```c
// Discard any locals created inside the loop.
for (int i = compiler->localCount - 1;
     i >= 0 && compiler->locals[i].depth > compiler->loop->scopeDepth;
     i--) {
    if(compiler->locals[i].isUpvalue)
    {
        emitByte(compiler, OP_CLOSE_UPVALUE);
    }
    else
    {
        emitByte(compiler, OP_POP);
    }
}
```

Jason2605:
Thanks for this!

Hmmm would be interesting to see an example script where this is causing issues - also what issues are you getting on windows?

Jason2605:
Ahhh very good find, thank you for providing the issue and the reproduction steps!! I'm more than happy to create the PR but if you want to get your name on the commits and into the repo open up that PR (even if it's just via the web portal and we can go from there!) - just let me know

RevengerWizard:
I'll try to open a pull request with the fix.

PR #609, opened 2023-03-07
Close upvalues after `continue` and `break` statements

RevengerWizard:

Correctly close upvalues after use of `break` and `continue` statements.

Resolves: #608 

### Type of change:

- [X] Bug fix


Jason2605:
Thanks for this !

PR #613, opened 2023-05-19
Pop switch expression  correctly

RevengerWizard:
This PR fixes a small behaviour bug with the switch statement. Prior to this, the switch expression was never popped out of stack at the end of the case statements, polluting the stack and leading to problems.

### Type of change:

- [X] Bug fix

### DIFF
--- a/src/vm/compiler.c
+++ b/src/vm/compiler.c
@@ -2350,8 +2350,8 @@ static void switchStatement(Compiler *compiler) {
 
     } while(match(compiler, TOKEN_CASE));
 
+    emitByte(compiler, OP_POP); // expression.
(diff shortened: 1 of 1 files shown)


RevengerWizard:
I think this could be more noticible when there is a switch statement inside a loop and taking a look at the traced instructions.
Here a basic example:

```js
for(var i = 1; i <= 3; i += 1)
{
    switch(i) 
    {
        case 1:
        {
            print("One");
            break;
        }
        case 2:
        {
            print("Two");
            break;
        }
        case 3:
        {
            print("Three");
            break;
        }
    }
    print("Other");
}
```

With a longer and intense loop it could be noticed better, but after all the compare jumps you should see the switch expression left up on the stack.

PR #614, opened 2023-05-29
Cmake preset

Ezbob:

Adding CMake preset support for setting up and building the interpreter. The main idea behind CMake presets is to store configurations of how CMake should be called, so that they can be easily shared between users of the project.

The presets added by this PR:

- `release` (both build and configuration) for release builds with HTTP support.
- `debug` (both build and configuration) for debug builds.
- `release-nohttp` (configuration only) configures a release build without HTTP support.

CMake presets, as they are used in this PR, requires CMake version 3.23.0 but you can still use the original, older, approach even if they are added (they are stored in a separate JSON file). 

More docs on CMake presets available from the [CMake preset documentation](https://cmake.org/cmake/help/v3.23/manual/cmake-presets.7.html)

### Type of change:

- [x] New way to build the project

#

### Housekeeping

#


Jason2605:
Nice, I like it - thank you!

PR #621, opened 2023-07-22
Add queue module

briandowns:

Added a module providing a queue.

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [X] New feature

#

### Housekeeping

#


briandowns:
Not really sure why CI is failing. Seems to exit without a message. Runs fine locally.

Jason2605:
Seems like the tests are failing for me locally, will try see what I can find

Jason2605:
Fixed the segfault but it seems some of the tests are failing / some memory isn't being cleared up, will continue looking

briandowns:
That's really annoying I'm not seeing that on my local. I'm happy to pull down the segfault fix and keep troubleshooting.

Jason2605:
Some comments as well about the actual API - I think providing a queue size could probably be optional and we resize it under the hood for the user (similar to how lists work) - what do you think?

I would probably also expect `.size()` (or maybe we provide a .len()) to give us the amount of items on the queue rather than the underlying size of malloc'd items, this would allow a user to iterate over a queue and pop as we need

src/optionals/queue.c
```diff
@@ -0,0 +1,203 @@
...
+
+    Queue *queue = AS_QUEUE(args[0]);
+
+    if (queue->cap != queue->count) {
+        if(queue->rear == queue->cap-1) {
+            queue->rear = -1;            
+        }
+
+        double *data = ALLOCATE(vm, double, 1);
+        *data = AS_NUMBER(args[1]);
```
Jason2605, line 104: Here we are assuming anything pushed to the queue is a number, but in reality we'd want this to be any Value type

briandowns: I forgot to update that. First go was to get numbers working. The weird thing is that all types work.

src/optionals/queue.c
```diff
@@ -0,0 +1,203 @@
...
+        return EMPTY_VAL;
+    }
+
+    if (!IS_NUMBER(args[0])) {
+        runtimeError(vm, "new() argument must be a numbers");
+        return EMPTY_VAL;
+    }
+
+    double cap = AS_NUMBER(args[0]);
+    if (cap == 0) {
```
Jason2605, line 168:
```suggestion
    if (cap <= 0) {
```

src/optionals/queue.c
```diff
@@ -0,0 +1,203 @@
...
+        return newResultError(vm, "error: capacity must be greater than 0");
+    }
+
+    ObjAbstract *abstract = newQueueObj(vm);
+    push(vm, OBJ_VAL(abstract));
+    
+    Queue *queue = abstract->data;
+    queue->dq = ALLOCATE(vm, void*, cap);
```
Jason2605, line 176: We can use Value rather than void*

briandowns:
Do we need to account for different architectures in tests? I'm seeing some really weird things in CI that don't quite make sense.

Jason2605:
> Do we need to account for different architectures in tests? I'm seeing some really weird things in CI that don't quite make sense.

Sorry somehow completely missed this comment? The tests for a queue module shouldn't need to be different per architecture no, it would only be when we use an architecture specific include I would have imagined. Are you meaning the failing tests in CI?

briandowns:
> > Do we need to account for different architectures in tests? I'm seeing some really weird things in CI that don't quite make sense.
> 
> Sorry somehow completely missed this comment? The tests for a queue module shouldn't need to be different per architecture no, it would only be when we use an architecture specific include I would have imagined. Are you meaning the failing tests in CI?

Yeah. Not sure where those errors are coming from. Of course, it works locally. :) Probably something dumb.

Jason2605:
That should be them passing now @briandowns, was just missing the ->capacity initialiser!

I'm curious too - 99% of the time I would imagine the end user not actually caring about the initial capacity size of the queue, so for this reason I suggest we:

- Have `.new()` be a default capacity size (probably 8?) then it means we don't need to worry about passing a parameter and it also means we don't need to worry about unwrapping the queue object (saves a little but of verbosity)
- Have a `.newWithSize()` (or similar name) which is essentially what `.new()` is now which will return a Result<Queue> type

Thoughts?

briandowns:
> That should be them passing now @briandowns, was just missing the ->capacity initialiser!
> 
> I'm curious too - 99% of the time I would imagine the end user not actually caring about the initial capacity size of the queue, so for this reason I suggest we:
> 
> * Have `.new()` be a default capacity size (probably 8?) then it means we don't need to worry about passing a parameter and it also means we don't need to worry about unwrapping the queue object (saves a little but of verbosity)
> * Have a `.newWithSize()` (or similar name) which is essentially what `.new()` is now which will return a Result<Queue> type
> 
> Thoughts?

Dig it. I'll get on it this evening. I don't know why this was working for me. Thank you!

Jason2605:
That's all good! Also seem to be getting some strange behaviour in the REPL ;(

```
Dictu Version: 0.27.0
>>> import Queue;
>>> const x = Queue.new(8).unwrap();
>>> x.push("test");
<Result Suc>
>>> const y = x.pop();
>>> print(y.unwrap());

>>> print(y.unwrap());
dictu(56734,0x11047fdc0) malloc: can't allocate region
:*** mach_vm_map(size=18446744073187381248, flags: 100) failed (error code=3)
dictu(56734,0x11047fdc0) malloc: *** set a breakpoint in malloc_error_break to debug
```

src/optionals/queue.c
```diff
@@ -0,0 +1,208 @@
...
+
+    Queue *queue = AS_QUEUE(args[0]);
+
+    if (queue->capacity < queue->count + 1) {
+        int oldCapacity = queue->capacity;
+        queue->capacity = GROW_CAPACITY(oldCapacity);
+        queue->dq = GROW_ARRAY(vm, queue->dq, Value, oldCapacity, queue->capacity);
+    }
+
+    if(queue->rear == queue->capacity-1) {
```
Jason2605, line 103:
```suggestion
    if (queue->rear == queue->capacity-1) {
```

src/optionals/queue.c
```diff
@@ -0,0 +1,208 @@
...
+        queue->rear = -1;            
+    }
+
+    queue->dq[++queue->rear] = args[1];
+    queue->count++;
+
+    return newResultSuccess(vm, NIL_VAL);
```
Jason2605, line 110: We shouldn't need the Result type here as there isn't an error state

```suggestion
    return NIL_VAL;
```

src/optionals/queue.c
```diff
@@ -0,0 +1,208 @@
...
+}
+
+static Value queuePop(DictuVM *vm, int argCount, Value *args) {
+    if (argCount != 0) {
+        runtimeError(vm, "pop() takes no arguments (%d given)", argCount);
+        return EMPTY_VAL;
+    }
+
+    Queue *queue = AS_QUEUE(args[0]);
+
```
Jason2605, line 120: We will want a check in here to make sure we're not attempting to pop on an empty queue

```suggestion
if (queue->count == 0) {
        runtimeError(vm, "pop() called on an empty queue");
        return EMPTY_VAL;
 }
```

src/optionals/queue.c
```diff
@@ -0,0 +1,208 @@
...
+    queue->count--;
+
+    if (queue->count < queue->capacity) {
+        int oldCapacity = queue->capacity;
+        queue->capacity = SHRINK_CAPACITY(oldCapacity);
+        queue->dq = SHRINK_ARRAY(vm, queue->dq, Value, oldCapacity, queue->capacity);
+    }
+
+    return newResultSuccess(vm, data);
```
Jason2605, line 135:
```suggestion

    return data;
```

We should definitely think about sizing down queue sizes if it meets a certain threshold, but we wouldn't want it to happen each call to `pop()` (as count is always going to be less than capacity in this case)

Something to think about with lists too actually, I don't think they size down

briandowns: Curious your thoughts on this. I changed the shrink operation to only shrink the underlying array if the count is less than half the capacity but greater than the default size.

Jason2605: I like it yep!

docs/docs/standard-lib/queue.md
```diff
@@ -0,0 +1,110 @@
...
+print(len);
+// 0
+```
+
+### Queue.isEmpty() -> Bool
+
+Returns a Bool value indicating whether the queue is empty or not.
+
+```cs
+const queue = Queue.new().unwrap();
```
Jason2605, line 95:
```suggestion
const queue = Queue.new();
```

docs/docs/standard-lib/queue.md
```diff
@@ -0,0 +1,110 @@
...
+print(res);
+// true
+```
+
+### Queue.isFull() -> Bool
+
+Returns a Bool value indicating whether the queue is empty or not.
+
+```cs
+const queue = Queue.new().unwrap();
```
Jason2605, line 106:
```suggestion
const queue = Queue.new();
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
+import Queue;
+
+{ // main
+    var queue = Queue.new(100).unwrap();
```
Jason2605, line 4:
```suggestion
    var queue = Queue.new();
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    queue.push(665);
+    print(queue.isEmpty());
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("peak: {}".format(queue.peek().unwrap()));
+    print(queue.isEmpty());
+
+    var q = Queue.new(100).unwrap();
```
Jason2605, line 21:
```suggestion
    var q = Queue.new();
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    queue.push(662);
+    queue.push(663);
+    queue.push(664);
+    queue.push(665);
+    print(queue.isEmpty());
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("pop: {}".format(queue.pop().unwrap()));
+    print("peak: {}".format(queue.peek().unwrap()));
```
Jason2605, line 18:
```suggestion
    print("pop: {}".format(queue.pop()));
    print("pop: {}".format(queue.pop()));
    print("pop: {}".format(queue.pop()));
    print("pop: {}".format(queue.pop()));
    print("peak: {}".format(queue.peek()));
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    q.push("1234");
+    q.push("0987");
+    q.push("poiu");
+    q.push("mnbv");
+    print(q.isEmpty());
+    print("pop: {}".format(q.pop().unwrap()));
+    print("pop: {}".format(q.pop().unwrap()));
+    print("pop: {}".format(q.pop().unwrap()));
+    print("pop: {}".format(q.pop().unwrap()));
+    print("peak: {}".format(q.peek().unwrap()));
```
Jason2605, line 35:
```suggestion
    print("pop: {}".format(q.pop()));
    print("pop: {}".format(q.pop()));
    print("pop: {}".format(q.pop()));
    print("pop: {}".format(q.pop()));
    print("peak: {}".format(q.peek()));
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    print(q.isEmpty());
+
+    class Person {
+        init(id="") {
+            this.id = id;
+        }
+    }
+
+    var q3 = Queue.new(100).unwrap();
```
Jason2605, line 44:
```suggestion
    var q3 = Queue.new();
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    q3.push(Person("2"));
+    q3.push(Person("3"));
+    q3.push(Person("4"));
+    q3.push(Person("5"));
+    print(q3.isEmpty());
+    print("pop: {}".format(q3.pop().unwrap().id));
+    print("pop: {}".format(q3.pop().unwrap().id));
+    print("pop: {}".format(q3.pop().unwrap().id));
+    print("pop: {}".format(q3.pop().unwrap().id));
+    print("peak: {}".format(q3.peek().unwrap().id));
```
Jason2605, line 58:
```suggestion
    print("pop: {}".format(q3.pop().id));
    print("pop: {}".format(q3.pop().id));
    print("pop: {}".format(q3.pop().id));
    print("pop: {}".format(q3.pop().id));
    print("peak: {}".format(q3.peek().id));
```

examples/queue.du
```diff
@@ -0,0 +1,62 @@
...
+    print(q3.isEmpty());
+    print("pop: {}".format(q3.pop().unwrap().id));
```
Jason2605, line 60:
```suggestion
    print("pop: {}".format(q3.pop().id));
```

docs/docs/standard-lib/queue.md
```diff
@@ -0,0 +1,110 @@
...
+
+Returns a Result value with the expected value on success or `Nil` on faiure. The capacity will be decreased as necessary.
+
+```cs
+const queue = Queue.new().unwrap();
+queue.push("some_value");
+const value = queue.pop();
+print(value);
+// some_value
+```
```
Jason2605, line 66: We will want to document the `.peek()` method as well

Jason2605 (commented):
Sorry last round of comments !

src/optionals/queue.c
```diff
@@ -0,0 +1,241 @@
...
+    queue->dq = ALLOCATE(vm, Value, DEFAULT_QUEUE_CAPACITY);
+    queue->capacity = DEFAULT_QUEUE_CAPACITY;
+    queue->front = 0;
+    queue->rear = -1;
+    queue->count = 0;
+
+    Value success = newResultSuccess(vm, OBJ_VAL(abstract));
+    pop(vm);
+
+    return success;
```
Jason2605, line 189: Sorry this one is on me for not being clear in the inital feedback, since we can't get an error state when creating a Queue object with `.new` we can just return the Queue directly, means we won't need to worry about unwrapping it

```suggestion
    pop(vm);

    return OBJ_VAL(abstract);
```

briandowns:
More HTTP failures. I'd restart but I don't have that option to do so in the interface.

Jason2605:
Thanks very much for this!

PR #625, opened 2023-08-14
Add hash Function to Object Module

briandowns:

Resolves: #

#

### Type of change:

- [ ] Bug fix

- [ ] New feature

#

### Housekeeping

#

### DIFF
--- a/src/optionals/object/object.c
+++ b/src/optionals/object/object.c
@@ -55,8 +55,8 @@ static Value objectHash(DictuVM *vm, int argCount, Value *args) {
     }
 
     char str[64];
-    sprintf(str, "%llu", args[0]);
-    return OBJ_VAL(copyString(vm, str, 64));
+    sprintf(str, "%llu", (unsigned long long)args[0]);
+    return OBJ_VAL(copyString(vm, (char *)str, 64));
 }
 
 Value createObjectModule(DictuVM *vm) {
--- a/src/optionals/object/object.c
+++ b/src/optionals/object/object.c
@@ -48,6 +48,17 @@ static Value objectGetClassRefInternal(DictuVM *vm, int argCount, Value *args) {
     return objectGetClassRefImpl(vm, argCount, args, true);
 }
 
+static Value objectHash(DictuVM *vm, int argCount, Value *args) {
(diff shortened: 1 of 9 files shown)


briandowns:
I started out returning the number but was problematic going from unsigned long long to double so resulted in returning the string form of the value.

src/optionals/object/object.c
```diff
@@ -48,6 +48,17 @@ static Value objectGetClassRefInternal(DictuVM *vm, int argCount, Value *args) {
...
 
+static Value objectHash(DictuVM *vm, int argCount, Value *args) {
+    if (argCount != 1) {
+        runtimeError(vm, "hash() takes 1 argument (%d given)", argCount);
+        return EMPTY_VAL;
+    }
+
+    char str[64];
+    sprintf(str, "%llu", (unsigned long long)args[0]);
+    return OBJ_VAL(copyString(vm, (char *)str, 64));
```
Jason2605, line 59:
```suggestion
    char str[21];
    snprintf(str, 21, "%llu", (unsigned long long)args[0]);
    return OBJ_VAL(copyString(vm, (char *)str, 20));
```

Although we're dealing with 64 bits the max size for a uint64_t is 18,446,744,073,709,551,615, so will be 20 chars

Jason2605:
I like it 👍 

I wonder too, with this, it's probably worth noting some of the caveats in the docs such as some Values will return the same hash regardless (e.g numbers / boolean / nil) and strings that will return the same hash at runtime since they are interned

briandowns:
I swear there's a special time of day that CI will pass successfully.  🤣

Jason2605:
> I swear there's a special time of day that CI will pass successfully. 🤣

Hahahah yeah it's becoming a pain, will look to try resolve that

briandowns:
I think it's more the httpbin side of things. I played with their API a little and saw the same timeouts. I even started a branch that swapped out the tests that call httpbin to another testing provider and then once into it found they didn't quite have thew same feature set. :/

Jason2605:
Yeah, something we can tackle, as long as the non-http tests pass I'm happy

Issue #623, opened 2023-08-09
[FEATURE] - Class Field Annotations


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It'd be really helpful for a few different projects I'm working on to have field level annotations.

An example: 

```js
abstract class Model {

    @PrimaryKey
    @AutoIncrement
    var id = 0;

    @Type("TIMESTAMP")
    var createdAt = "";

    @Type("TIMESTAMP")
    var updatedAt = "";

    @Type("TIMESTAMP")
    var deletedAt = "";

    // used for processing
    var usingModel = true;
}
```

### Describe alternatives you've considered

N/A

Jason2605:
I have a branch on `feature/attributeAnnotations` that should have this in place now if you want to check it out. Need to write some tests and some docs for it but the functionality should be in place. Will work the same as the other annotations except it is exposed on `attributeAnnotations` (potentially going to change this)

briandowns:
That's awesome! Happy to check it out and do some testing.

Jason2605:
Nice one, thank you! Out of curiosity what are you making? Some kind of ORM?

briandowns:
No prob. As for what I'm making, exactly, a very simple, albeit, naive ORM. Should be interesting.

Jason2605:
That sounds really cool! Hopefully the feature branch works for you then

PR #627, opened 2023-08-14
Class Variable / Constant Annotations

Jason2605:

This adds a new `.variableAnnotations` to classes that exposes any attached annotations to either class variables or class constants. Works exactly the same way as the other annotations where a dictionary is generated.

Resolves: #623 

#

### Type of change:

- [x] New feature

#

### Housekeeping

#


Jason2605:
@briandowns This should be working now - not sure I'm happy with `variableAnnotations` when it is also used for constants (contradicting a bit), do you have an idea for a better name?

briandowns:
All of it looks great and I can't wait to finish the 2 modules I have in mind but as for the method name, I think I agree. How about `fieldAnnotations`?

Jason2605:
Thanks man 😄 Yeah I like it, will change it to that

docs/docs/classes.md
```diff
@@ -649,7 +649,7 @@ anotherTestObj.isInstance(Test); // true
 
 ## Annotations
 
-Annotations are metadata that are applied to classes and methods that by themselves have no impact. They, however, can provide user defined changes at runtime.
+Annotations are metadata that are applied to classes, methods and class variables that by themselves have no impact. They, however, can provide user defined changes at runtime.
```
briandowns, line 652:
```suggestion
Annotations are metadata that are applied to classes, methods and class variables and constants that by themselves have no impact. They, however, can provide user defined changes at runtime.
```

Issue #637, opened 2023-08-21
[BUG] - Unable to Create class, var, or const with the Name "C"


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I ran into an issue while testing other code when I tried to create a class named "C". I think it's conflicting with the internal module of the same name. I tested with const's and var's and it's the same behavior.

### Steps To Reproduce

Using the latest version:

```sh
>>> class C {}
>>> const c = C();
Error: File 'repl', line 1
       'module' is not callable
```

```sh
>>> const C = 90;
>>> print(C);
<Module C>
>>> var C = 90;
>>> print(C);
<Module C>
```

briandowns:
I adjusted the code to not auto load the C module but to require it be explicitly imported, tested, and seemed to work fine. I can push a PR but wanted to get your (@Jason2605) thoughts and feedback on the matter before I did.

Jason2605:
Please yeah, I'm unsure the reason I didn't autoload that module

PR #638, opened 2023-08-21
un-auto load C module

briandowns:
Resolves: #637 

### What's Changed:

Updated the code to require the C module to be loaded like all other modules and not be autoloaded with the VM.

#

### Type of Change:

- [x] Bug fix
- [ ] New feature

#

### Housekeeping:

#


briandowns:
@Jason2605 Here's the initial pass at the fix. Happy to add docs if necessary. Not really much there to test though but happy to add some.

Jason2605:
To be honest for the moment I'm happy with no tests / docs on this module, at the moment it's really just a bunch of constants since a lot got stripped out when we moved to Result types

Jason2605:
Thanks for this!

PR #642, opened 2023-08-28
Fix issue with string operations on less and greater than

Jason2605:
# Operators

### What's Changed:

Fixes an issue with the less than and greater than operators on strings

#

### Type of Change:

- [x] Bug fix
- [ ] New feature

#

### Housekeeping:

#

### DIFF
--- a/src/vm/value.c
+++ b/src/vm/value.c
@@ -577,6 +577,8 @@ bool compareStringLess(Value a, Value b) {
     for (i = 0; firstString->chars[i] != '\0' || secondString->chars[i] != '\0'; i++) {
         if (firstString->chars[i] < secondString->chars[i]) {
             return true;
+        } else if (firstString->chars[i] > secondString->chars[i]) {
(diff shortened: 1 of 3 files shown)


src/vm/value.c
```diff
@@ -577,6 +577,8 @@ bool compareStringLess(Value a, Value b) {
     for (i = 0; firstString->chars[i] != '\0' || secondString->chars[i] != '\0'; i++) {
         if (firstString->chars[i] < secondString->chars[i]) {
             return true;
+        } else if (firstString->chars[i] > secondString->chars[i]) {
```
briandowns, line 580: Not sure how specific you like things but I don't believe the `else` here and below are necessary.

Jason2605: Yeah I guess you're right since we're returning early the else keyword is a bit redundant. Will change that

briandowns (approved):
LGTM!

Issue #630, opened 2023-08-16
[FEATURE] Mocking


Jason2605:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It would be good to extend the UnitTest module so that we can create mock objects, and then from that do things such as check certain methods were called, number of times they were called, arguments they were called with etc.

PR #644, opened 2023-08-28
Mocking

Jason2605:
# Mocking

Resolves: #630 

### What's Changed:

Adds a new `mock()` function to the UnitTest module that will allow us to mock certain Values. In the current implementation it will allow us to mock a given class and also return given values on certain method calls.

If no return values are specified via the second argument dictionary then it will implicitly create an empty method that returns nil.

Example script:
```
from UnitTest import mock;

class Base {
    base() {
        print("hello!");
    }
}

class Test < Base {
    test() {
        print('hello');
    }

    someFunc() {
        return 100;
    }
}

const mockObj = mock(Test, {
    "test": [1, 2, 3]
});

print(mockObj.base()); // nil
print(mockObj.test()); // [1, 2, 3]
print(mockObj.someFunc()); // nil
```

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


Jason2605:
@briandowns before going much further with this one I was thinking what you think of this ^?

Jason2605:
@briandowns Do the stack tests fail for you locally? I'm not seeing this happen and I'm currently unsure why

Jason2605:
> @briandowns Do the stack tests fail for you locally? I'm not seeing this happen and I'm currently unsure why

Ah okay I got it, there is a problem with both the stack and queue modules. We are storing references to values within the abstract data type that will never get marked by the GC

briandowns:
Didn't see anything weird locally. My bad on missing the GC.

Jason2605:
It's all good, it's quite a nasty issue actually! None of the abstracts have a way of graying any values so was a needed change. I also didn't see anything locally, but I noticed the tests are mainly using numbers, if we change a test to say:

```
testStackShrink() {
        this.assertEquals(this.stack.cap(), this.defaultStackSize);
        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16].forEach(def (x) => this.stack.push(x.toString()));
        this.assertEquals(this.stack.cap(), 32);
        const res = this.stack.pop();
        this.assertEquals(res, 16);
        this.assertEquals(this.stack.cap(), 32);
        this.stack.pop();
        this.stack.pop();
        this.stack.pop();
        this.stack.pop();
        this.stack.pop();
        this.stack.pop();
        this.stack.pop();
        this.assertEquals(this.stack.cap(), 16);
    }
```

And we have objects that we actually get wiped by the GC I see the segfault come through

Issue #633, opened 2023-08-18
[FEATURE] - Classes should inherit annotations from inherited classes


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It's be nice to have field, and method annotations brought along when a class is inherited from another.

PR #654, opened 2023-09-03
Add Annotation Inheritance, tests, and docs

briandowns:
Resolves: #633

### What's Changed:

Added the ability to have classes inherit class, method, and field annotations.

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


src/vm/vm.c
```diff
@@ -796,6 +811,20 @@ static void setReplVar(DictuVM *vm, Value value) {
...
+    for (int i = 0; i <= superAnnotations->capacityMask; ++i) {
+        if (IS_EMPTY(superAnnotations->entries[i].key)) {
+            continue;
+        }
+
+        Value val = superAnnotations->entries[i].value;
+        push(vm, val);
+
+        dictSet(vm, klassAnnotations, superAnnotations->entries[i].key, val);
+        pop(vm);
```
Jason2605, line 824: We shouldn't need to push to the stack here as they should already be referenced in the parent class, which means they will get greyed there

src/vm/vm.c
```diff
@@ -2187,6 +2216,20 @@ static DictuInterpretResult run(DictuVM *vm) {
             ObjDict *dict = AS_DICT(READ_CONSTANT());
             ObjClass *klass = AS_CLASS(peek(vm, 0));
 
+            if (dict->entries == 0) {
+                printf("XXX - 0!\n");
+            }
```
Jason2605, line 2221:
```suggestion
```

Jason2605:
Couple of things we will need to think about with this when testing actually - the first is if the child class overrides something in the parent class we want to take the childs implementation not the parents

```
class Base {
    @method
    test() {

    }
}

class Test < Base {
    @method1234
    test() {

    }
}

print(Test.methodAnnotations); // {"test": {"method": nil}}

// Where I would expect {"test": {"method1234": nil}}
```

I would imagine some sort of key check in copyAnnotations should do the trick.

Similarly (not sure why this would happen) but if we modify the annotation value on the child we want to make sure it doesn't update the parent (essentially we would want a deep copy)

```
class Base {
    @method
    test() {

    }
}

class Test < Base {

}

Test.methodAnnotations['test']['method'] = 10;

print(Test.methodAnnotations); // {"test": {"method": 10}}
print(Base.methodAnnotations); // {"test": {"method": 10}}
```

briandowns:
Great catches. Impl updated as well as tests.

src/vm/vm.c
```diff
@@ -761,6 +761,21 @@ static void createClass(DictuVM *vm, ObjString *name, ObjClass *superclass, Clas
     if (superclass != NULL) {
         tableAddAll(vm, &superclass->publicMethods, &klass->publicMethods);
         tableAddAll(vm, &superclass->abstractMethods, &klass->abstractMethods);
+
+        if (superclass->classAnnotations != NULL) {
+            ObjDict *dict = copyDict(vm, superclass->classAnnotations, true);
```
Jason2605, line 766:
```suggestion
            ObjDict *dict = copyDict(vm, superclass->classAnnotations, false);
```

src/vm/vm.c
```diff
@@ -761,6 +761,21 @@ static void createClass(DictuVM *vm, ObjString *name, ObjClass *superclass, Clas
...
+            klass->classAnnotations = dict;
+        }
+
+        if (superclass->methodAnnotations != NULL) {
+            ObjDict *dict = copyDict(vm, superclass->methodAnnotations, true);
```
Jason2605, line 771:
```suggestion
            ObjDict *dict = copyDict(vm, superclass->methodAnnotations, false);
```

src/vm/vm.c
```diff
@@ -761,6 +761,21 @@ static void createClass(DictuVM *vm, ObjString *name, ObjClass *superclass, Clas
...
+            klass->methodAnnotations = dict;
+        }
+
+        if (superclass->fieldAnnotations != NULL) {
+            ObjDict *dict = copyDict(vm, superclass->fieldAnnotations, true);
```
Jason2605, line 776:
```suggestion
            ObjDict *dict = copyDict(vm, superclass->fieldAnnotations, false);
```

src/vm/vm.c
```diff
@@ -2187,6 +2219,16 @@ static DictuInterpretResult run(DictuVM *vm) {
...
 
+            if (IS_CLASS(peek(vm, 1))) {
+                ObjClass *super = AS_CLASS(peek(vm, 1));
+
+                if (super->classAnnotations != NULL) {
+                    ObjDict *superAnnotations = copyDict(vm, super->classAnnotations, true);
+
+                    copyAnnotations(vm, superAnnotations, dict);
+                }
+            }
```
Jason2605, line 2230: I just kinda had a realisation with this bit of code as well (and the other two opcodes)

We can simply just do:
```suggestion
            if (klass->classAnnotations != NULL) {
                copyAnnotations(vm, klass->classAnnotations, dict);
            }
```

As the dict on the class has already been copied from the parent so we should be free to modify this directly

Jason2605 (commented):
Sorry couple more comments!

briandowns:
All set!

Jason2605:
Nicceeee thanks for this!!

briandowns:
No prob. Thanks for the quick reviews! My ORM is really coming along now that all of the annotation features are in!

Jason2605:
Awesome! Definitely be interested to see it! Also side note, I ended up using Wanbli for a little side project of mine :)

briandowns:
Open to any and all feedback on that. I don't really have it in a stable state yet so use with caution!

Jason2605:
It’s all good just a little tool I’m using in the house to track some smart plug usage! I’ll potentially PR some things across for it but it’s been a fun project

briandowns:
Here's the ORM. Suuuuper naive... Not finished but works. https://github.com/briandowns/kahless . Gonna get an example together using Kahless and Wanbli together.

PR #656, opened 2023-09-06
Class to dict

briandowns:
Resolves: #

### What's Changed:

Adds the ability to convert a class to a dictionary.

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


briandowns:
I see the error messages but kinda at a loss as this is passing locally with no issue.

Jason2605:
Pushed up a couple of changes there:
- Walk the full inheritance chain rather than just one parent up
- Fix a couple of memory issues where the strings weren't being pushed and the dicts/lists were being popped too early
- Add a test case for inheritance

Jason2605:
Hahaha yeah 4 years down the line I'm still wrapping my head around it! Thanks for this

Issue #647, opened 2023-08-29
[FEATURE] Datetime


Jason2605:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Overhaul the Datetime built in module so that it's a lot more feature rich. Rather than simply wrapping the C functions it would be nice if dealt with an actual Datetime object being returned that we can manipulate: e.g addDay, subDay, etc

MannarAmuthan:
Hi , Could I take this one ?

MannarAmuthan:
Hi, Could I take this one?

Jason2605:
Of course!

MannarAmuthan:
Raised #675

PR #675, opened 2023-10-02
[Feature] Implemented Datetime object support

MannarAmuthan:
# Implemented Datetime object

Resolves: #647 

### What's Changed:
I didn't removed old contracts yet, means there is no any breaking changes, confirmed with tests. Added instantiation for datetime objects, once we agreed on these contracts, other utility methods for this datetime objects can be implemented on top of it. 

### Type of Change:

- [ ] Bug fix
- [x ] New feature

#

### Housekeeping:

- [ x] Tests have been updated to reflect the changes done within this PR (if applicable).
- [ x] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

#### Constructors
1. Datetime.new() -> Datetime
2. Datetime.new(String, String) -> Datetime
3. Datetime.new(Number) -> Datetime
#### Methods
1. datetimeObj.strptime() -> Number
2. datetimeObj.strftime(String) -> String

### DIFF
--- a/docs/docs/standard-lib/datetime.md
+++ b/docs/docs/standard-lib/datetime.md
@@ -149,7 +149,7 @@ const datetime = Datetime.new("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00");
 datetime.strptime(); // 1577836800
 ```
 
-### datetimeObj.strftime() -> String
+### datetimeObj.strftime(String) -> String
 
 Returns a user-defined datetime formatted string for a datetime object, see [Datetime formats](#datetime-formats).
 
--- a/docs/docs/standard-lib/datetime.md
+++ b/docs/docs/standard-lib/datetime.md
@@ -142,3 +142,50 @@ const datetime = Datetime.new();
 datetime.strftime(); // Sat Oct 21 21:44:25 2023
 ```
 
+
+## Duration
+
+A Duration object represents a time delta, the difference between two datetimes. This Duration object can be used to create new datetime objects, by performing operations such as addition and subtraction.
+
+### Datetime.duration(Dict) -> Duration
+
+Duration object is constructed by passing dictionary duration properties. weeks, days, hours, minutes and seconds are allowed properties.
+
+```cs
+import Datetime;
+const duration = Datetime.duration({'weeks':0, 'days': 1, 'seconds': 0, 'minutes': 0, 'hours': 0});
+```
+
+### datetimeObj.add(Duration) -> Datetime
+
+Returns a new Datetime object after adding given duration.
+
+```cs
+const duration = Datetime.duration({'weeks':0, 'days': 1, 'seconds': 0, 'minutes': 0, 'hours': 0});
+const datetime = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00");
+datetime.add(duration).strftime(); // Thu Jan 02 00:00:00 2020
+```
+
+
+### datetimeObj.sub(Duration) -> Datetime
+
+Returns a new Datetime object after subtracting given duration.
+
+```cs
+const duration = Datetime.duration({'weeks':53});
+const datetime = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00");
+datetime.sub(duration).strftime(); // Wed Jan 02 00:00:00 2019
+```
+
+### datetimeObj.delta(datetimeObj) -> Duration
+
+Returns a Duration object represents a delta with a given date.
+
+```cs
+var datetimeOne = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00");
+var datetimeTwo = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-02 00:00:00");
+var duration = datetimeOne.delta(datetimeTwo); 
+
+const datetime = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-05-01 00:00:00");
+datetime.add(duration).strftime(); // Sat May 02 00:00:00 2020
+```
--- a/docs/docs/standard-lib/datetime.md
+++ b/docs/docs/standard-lib/datetime.md
@@ -140,6 +140,8 @@ Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00"); // 1577836800
 
 Returns a number which is the number of seconds from epoch, for a given datetime
 
+**Note:** This is not available on windows systems.
+
 ```cs
 const datetime = Datetime.new("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00");
 datetime.strptime(); // 1577836800
--- a/docs/docs/standard-lib/datetime.md
+++ b/docs/docs/standard-lib/datetime.md
@@ -40,18 +40,7 @@ Returns a datetime object with current datetime.
 
 ```cs
 var datetime = Datetime.new(); 
-datetime.toString(); // Fri May 29 03:12:32 2020
-```
-
-### Datetime.new(String, String) -> Datetime
-
-Returns a datetime object for given time string format.
-
-**Note:** This is not available on windows systems.
-
-```cs
-var datetime = Datetime.new("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00"); 
-datetime.toString(); // Wed Jan  1 00:00:00 2020
+datetime.strftime(); // Fri May 29 03:12:32 2020
 ```
 
 ### Datetime.new(Number) -> Datetime
@@ -60,7 +49,7 @@ Returns a datetime object for given number of seconds from epoch.
 
 ```cs
 const datetime = Datetime.new(1577836800);
-datetime.toString(); // Wed Jan  1 00:00:00 2020
+datetime.strftime(); // Wed Jan  1 00:00:00 2020
 ```
 
 
@@ -78,6 +67,18 @@ Returns a human readable UTC datetime string.
 
 ```cs
 Datetime.now(); // Fri May 29 02:12:32 2020
+```
+
+
+### Datetime.strptime(String, String) -> Datetime
+
+Returns a datetime object for given time string format.
+
+**Note:** This is not available on windows systems.
+
+```cs
+var datetime = Datetime.strptime("%Y-%m-%d %H:%M:%S", "2020-01-01 00:00:00"); 
+datetime.strftime(); // Wed Jan  1 00:00:00 2020
 ```
 
 ### Datetime formats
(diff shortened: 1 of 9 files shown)


MannarAmuthan:
Looks like there is some issue in handling gmt, will fix it

src/optionals/datetime.c
```diff
@@ -163,8 +397,14 @@ Value createDatetimeModule(DictuVM *vm) {
...
      * Define Datetime methods
      */
+    
+    defineNative(vm, &module->values, "new", newDatetimeNative);
+    defineNative(vm, &module->values, "newUTC", newUTCDatetimeNative);
     defineNative(vm, &module->values, "now", nowNative);
     defineNative(vm, &module->values, "nowUTC", nowUTCNative);
+
+    // defineNative(vm, &module->values, "now", nowNative);
+    // defineNative(vm, &module->values, "nowUTC", nowUTCNative);
```
Jason2605, line 407: If these arent needed we can remove

```suggestion

```

tests/datetime/constants.du
```diff
@@ -15,30 +15,30 @@ class TestDatetimeConstants < UnitTest {
     const DATE_TIME_FORMAT = "%a %b %d %H:%M:%S %Y";
 
     testSecondsInMinuteConstant() {
-        const startSec = Datetime.strptime(this.DATE_TIME_FORMAT, "Fri May 29 03:12:32 2020");
+        const startSec = Datetime.new(this.DATE_TIME_FORMAT, "Fri May 29 03:12:32 2020").strptime();
```
Jason2605, line 18: For these can we leave the existing tests alone but also add in new tests?

Jason2605 (commented):
Couple of minor points

Jason2605:
Appreciate the work you've done on this so i've stuck the hacktoberfest label on it (if you're participating) until I get some time to thoroughly check it and get it merged in!

Jason2605:
I think I'm going to work on this PR (unless you want to continue with it) so that all of the object methods return a Datetime object rather than primitive values and have extra methods on the datetime object to convert to string / numeric values.

Means as we add more methods we can much easier chain things (e.g like addHour or whatever)

MannarAmuthan:
Certainly, could you provide some contract ideas that I can attempt to implement ?.
Because current methods such as datetimeObj.strptime,  datetimeObj.strftime could not return Datetime object. Am I right?

Jason2605:
Exactly yeah, we'd leave the current ones in and they can return the primitive values. Your new object methods would instead return a Datetime object, so:

```
Datetime.new(<values>); -> Datetime
Datetime.strptime(<string>, <string>); -> Datetime
Datetime.getTime(); -> Number
Datetime.strftime(<string>); -> String
Datetime.toString(); -> String (thinking do we need this if we have strftime actually? we could have this be strftime with no format specifier passed to it)


// Later down the line
Datetime.addMinutes(<number>); -> Datetime
Datetime.addHours(<number>); -> Datetime
...
```

MannarAmuthan:
Thanks for Suggestions,
I assume that we have two different types of methods/functions, one is on Datetime module, and another is on Datetime object itself.

Methods of Datetime module
=====================
Already there
--------------
Datetime.new(); -> Datetime(Object) , current time object
Datetime.newUTC(); -> Datetime(Object) , current time object

To implement/modify
--------------
Datetime.new(<values>); -> Datetime (Object)
Does values means day,month,year, hour, etc ..?
Datetime.strptime(<string>, <string>); -> Datetime(Object) , currently it returns string i guess

Methods of Datetime Object
=====================
Already there
--------------
dateTime(object).strftime(<string>)  -> String , returns string

In future
---------
dateTime(object).addMinutes(<number>); -> Datetime(Object)
dateTime(object).addHours(<number>); -> Datetime(Object)
...

Jason2605:
Apologies yeah, step time to return a date time object rather than a number, and then getTime to return a number

I had <values> because at the moment this is overloaded, that’s fine by me and can stay as is

Jason2605:
No rush or anything @MannarAmuthan but just checking if you're still working on this?

MannarAmuthan:
Hi yes , I am planning to work on this in weekend

MannarAmuthan:
Hi @Jason2605 , Please look into it, and provide any suggestions or changes needed.

Jason2605:
Will take a look later today, thanks @MannarAmuthan

docs/docs/standard-lib/datetime.md
```diff
@@ -82,28 +114,31 @@ Datetime.now(); // Fri May 29 02:12:32 2020
...
-the first parameter being the date format, see [Datetime formats](#datetime-formats) and the second
-the date string in the format of the first parameter.
+Returns a user-defined datetime formatted string for a datetime object, see [Datetime formats](#datetime-formats).
+
+```cs
+const datetime = Datetime.new();
+datetime.strftime("Today is %A"); // Today is Friday
+```
+
+Returns in default format when user defined format is not provided
```
Jason2605, line 137: Be nice to note the default format here: `%a %b %d %H:%M:%S %Y`

src/optionals/datetime.c
```diff
@@ -11,88 +11,78 @@
...
-        runtimeError(vm, "nowUTC() takes no arguments (%d given)", argCount);
-        return EMPTY_VAL;
-    }
+char *datetimeTypeToString(ObjAbstract *abstract) {
+    UNUSED(abstract);
 
-    time_t t = time(NULL);
-    struct tm tictoc;
-    char time[26];
+    char *queueString = malloc(sizeof(char) * 11);
```
Jason2605, line 30:
```suggestion
    char *datetimeString = malloc(sizeof(char) * 11);
```

Jason2605: *And same below

src/optionals/datetime.c
```diff
@@ -11,88 +11,78 @@
...
 
-    // -1 to remove newline
-    return OBJ_VAL(copyString(vm, time, strlen(time) - 1));
-}
 
-static Value strftimeNative(DictuVM *vm, int argCount, Value *args) {
-    if (argCount != 1 && argCount != 2) {
-        runtimeError(vm, "strftime() takes 1 or 2 arguments (%d given)", argCount);
+    if (argCount > 1) {
+        runtimeError(vm, "strftime() takes 0 or 1 argument (%d given)", argCount);
```
Jason2605, line 39:
```suggestion
        runtimeError(vm, "strftime() takes 0 or 1 arguments (%d given)", argCount);
```

Jason2605 (commented):
Thanks for this!! Few small comments really

Jason2605:
@briandowns Be good to get your thoughts on this too

briandowns:
I like this. 

I'm curious if we could simplify it though.

I was looking at working on a time module myself at one point and was going to model it after the Time package in Go. Essentially, the Time type is just a number and operations are performed on that value(s). There's a Duration type as well which is just an alias for an int64. The package's API is small and clean as well. https://cs.opensource.google/go/go/+/master:src/time/time.go;drc=bc2124dab14fa292e18df2937037d782f7868635;l=135

MannarAmuthan:
Hi @briandowns , That one is great reference, Thanks for this !

I assume we are talking about the structure of datetime object , will take look into it.

src/optionals/datetime.c
MannarAmuthan: These suggestions are done

Jason2605:
> I like this.
> 
> I'm curious if we could simplify it though.
> 
> I was looking at working on a time module myself at one point and was going to model it after the Time package in Go. Essentially, the Time type is just a number and operations are performed on that value(s). There's a Duration type as well which is just an alias for an int64. The package's API is small and clean as well. https://cs.opensource.google/go/go/+/master:src/time/time.go;drc=bc2124dab14fa292e18df2937037d782f7868635;l=135

Yeah this is an interesting concept, will need to have a deeper look through the Go source but I think representing the time as a number internally makes a lot of sense to me

MannarAmuthan:
Actually I read more on this, and found that classic **_time_t_** epoch integer, represents time after 1970. I changed the code, it is working fine, 

**New structure:**

typedef struct {
    long time;    /* timestamp integer*/
    int is_local; 
} Datetime;

But I worry, that we need the capability to create Datetime before 1970 also. 
If you have any thoughts on this let me know.

briandowns:
As always, I'll defer to @Jason2605 but I think we can be fine with storing a wall time and mono time in int64_t values in a dateTime object. From there, that single object can be used for timers, time diffs, and we can cast up and down from time_t. In the case of timestamps prior to epoch, they end up being negative numbers.

Jason2605:
> As always, I'll defer to @Jason2605 but I think we can be fine with storing a wall time and mono time in int64_t values in a dateTime object. From there, that single object can be used for timers, time diffs, and we can cast up and down from time_t. In the case of timestamps prior to epoch, they end up being negative numbers.

Honestly for this one, I'd probably go with whatever the go library is doing (still need to properly look at it sorry). A single numeric value storing the current time sounds perfect to me and going into negative numbers for dates prior to 1970 is fine (and what happens when you want a time before then when using unix time anyways yeah)

briandowns:
A simple and naive assumption of how this might look:

```c
typedef struct {
    int64_t wall;
    int64_t mono;
} DateTime;

DateTime now() {
    struct timespec wall, mono;

    clock_gettime(CLOCK_MONOTONIC, &mono);
    clock_gettime(CLOCK_REALTIME, &wall);

    DateTime dt;
    dt.mono = mono.tv_sec;
    dt.wall = wall.tv_sec;

    return dt;
}
```

Rooted off of the DateTime class, we can any number of methods that will use the values therein. Timers using mono, time and date operations using wall.

MannarAmuthan:
Thanks for all suggestions, I am planning to work on this , will provide updates

MannarAmuthan:
Currently, I have created a datetime object to store time as an integer and included a flag for location (local or not). However, this is not quite complete; I need to implement ADD and SUB methods  ( and duration ) to ensure proper functionality. I will investigate this further and continue working on it.

Jason2605:
Thanks for still working on this! Will be a really nice addition to get a more fully featured datetime library

MannarAmuthan:
## Suggestion and Approaches for introducing timedelta (or) duration object:

**Naming:**  The name **_timedelta_** is inspired from python datetime package , meanwhile **_duration_** also makes sense

**Note:** These are just initial thoughts, welcom more suggestions, even hybrid approaches.

**To Remember:** 
In Dictu, there is no named or keyword arguments for function calls, which restrict us to defining clear contract for timedelta object.

### Approach 1: Constructing TimeDelta by dictionary

```
mydatetime = Datetime.new();

toAdd = Timedelta({
'days': 2,
'hours': 1
});  // Can have year , month, etc

mydatetime.add(toAdd); // Returns new datetime object
```

#### Pros

- Can have more operations such as getting delta difference of two datetime objects, as delta object, etc.
- Having clear difference with datetime object

#### Cons
- API might looks slightly complex, especially introducing new object.
- Constrcution of timedelta object looks little nasty


### Approach 2: Avoid delta object

```
mydatetime = Datetime.new();
mydatetime.addDays(1);
mydatetime.addMinutes(2);
mydatetime.addYears(1); 
//etc
```

or 

```
mydatetime = Datetime.new().addDays(1).addMinutes(2).subSeconds(1);
```

#### Pros

- Very simple and minimal in terms of design

#### Cons

- Limited functionality 
- Might lead to defining lot of methods

Jason2605:
Thanks very much for the detailed write up! I personally thing the Timedelta option is the way to go - I feel like when dealing with calculating time differences it will be a lot easier to conditionally create a timedelta object than it will be to apply different methods - it also means we will be able to have an actual value representation of a length of time (we can then apply nice string methods to this for example)!

No kwargs is a valid gripe and would be a really nice addition, but as you say for now, a dictionary would be the next best replacement for instantiation

Be good to hear your thoughts too @briandowns

Jason2605:
Yeah when you run `.add` or `.minus` (or whatever we decide the public API to be) I would expect a new datetime object yeah!

I'm not too fussed on Duration vs Timedelta either, Luxon a popular JS library seems to use Duration whereas Python's implementation as mentioned uses Timedelta, so open to either on the naming front.

Comparisons of objects will be an interesting one in the sense of would you expect the standard operators to work with these types (< > != ==) or would we have comparison methods (potentially opens up the question of user defined operator overloading too but can save that for another day)

briandowns:
I'm not too concerned around naming Go uses an int64 as type Duration. Meh. 

As for object comparison, I'd probably at least expect a couple methods like t2.equals(t1);, and maybe a diff method.

Jason2605:
So you'd prefer methods rather than overloading the operators?

Makes sense, yeah I think im siding towards `Duration` at the moment

briandowns:
Yeah, for sure for methods over operators. Seems like the surface area for that is super small and still achieves the necessary functionality.

briandowns:
If this has stagnated, I'd like to pick it up.

Jason2605:
> If this has stagnated, I'd like to pick it up.

Have at it mate!

MannarAmuthan:
Sorry for dropping this earlier, but I'd love to jump back in and help now. Let me know if there are any updates or changes needed from my side, and I'll get on it as soon as possible.

I guess, I've finished most of the things, but I am not sure about the failures. Need to check

Jason2605:
Closing in favour of #750

PR #677, opened 2023-10-09
Add negate method to number type

briandowns:
Add negate method to number type

Resolves: #

### What's Changed:

Adds a new method, `negate`, to the number type.

### Type of Change:

- [ ] Bug fix
- [X] New feature

#

### Housekeeping:

- [X] Tests have been updated to reflect the changes done within this PR (if applicable).
- [X] Documentation has been updated to reflect the changes done within this PR (if applicable).

#

### DIFF
--- a/docs/docs/variables.md
+++ b/docs/docs/variables.md
@@ -16,16 +16,16 @@ nav_order: 3
 ---
 ## Data types
 
-| Type | Example | Note |
-| --- | --- | --- |
-| String | `'foo'`, `"bar"` | You can use single or double quotes to represent strings in Dictu. |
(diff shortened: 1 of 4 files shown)


Jason2605:
Hmm interesting, we can use `-` as a prefix operator rather than a function call for negating a number - do you think we need this as a method?

```
const x = 100;
print(-x); // - 100
```

briandowns:
Probably unnecessary then. I'll defer to you if you want it in the language. I'm happy to close the PR if you don't. 😀

Jason2605:
Appreciate the PR but I think we probably don’t need it

briandowns:
Cool, cool. Closing.

Issue #678, opened 2023-10-14
Contributing md file


amitkumar-y:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Having a Contributing.md file help the new contributors to understand the process of how to contribute to this project.

### Describe the solution you'd like

I think the project should have a contributing.md file.
If it sounds good then you can assign this to me I will add the file.

amitkumar-y:
Hi @Jason2605 ,
I have added the Contributing.md file [here](https://github.com/dictu-lang/Dictu/pull/681). Let me know if it looks good or if you want me to make any changes.

PR #681, opened 2023-10-20
Add Contributing.md file

amitkumar-y:
# Add the Contributing.md file to the repository.

Resolves: #678 

### What's Changed:

New details for starting the contribution to this project have been added in the Contributing.md file. It will assist those who wish to contribute.
#

### Type of Change:

- [ ] Bug fix
- [x ] New feature

#

### Housekeeping:

#

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,68 @@
+# How to contribute to the project
+
+This project welcomes any change, improvement, or suggestion!
+
+If you'd like to help its development feel free to open a new issue and raise a pull request.
+
+## IMPORTANT
+
+If you'd like to work on an existing issue, kindly **ask** for it to be assigned to you.
+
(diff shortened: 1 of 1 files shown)


Jason2605:
Thanks for this but we potentially need to think about what to put in this file. I feel like making it a guide on how to make a PR probably isn't what we want, and instead we want something maybe a little more in-depth on how to contribute to the source.

I think things like how to modules work in the VM, how we can add more native functions, how we can add functions to types etc are all things that are definitely worth documenting somewhere (maybe a wiki?) then that can be referenced in this file.

For your effort i'll stick a hacktoberfest accepted on it, sorry probably should have went through it more on the issue

briandowns:
I agree with the synopsis above. The contributors doc should be the guide anyone can use to update tests, docs, modules, etc.

PR #684, opened 2023-10-23
Parse Hex

Jason2605:
# Parse Hex

### What's Changed:

Ability to parse hex codes

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


src/vm/compiler.c
```diff
@@ -1037,6 +1037,28 @@ int parseEscapeSequences(char *string, int length) {
                     string[i + 1] = '\v';
                     break;
                 }
+                case 'x': {
+                    if (string[i + 2] == '\0' || string[i + 3] == '\0') {
+                        error(parser, "\\x escape code expects format \\xhh");
+                        break;
+                    }
+
```
briandowns, line 1045: Can we add an additional check here to make sure the given characters are in fact valid hex (0123456789abcdefABCDEF)?

Jason2605: That should happen below in the strtol check, but potentially an easier check is checking for the char being in that range

briandowns: Oh my bad, I just noticed that now. I'll defer to you. :)

Jason2605: No worries! Thanks for double checking anyways

PR #686, opened 2023-10-30
Add net module, tests, and docs

briandowns:
# Add net module, tests, and docs

### What's Changed:

Adds a new module called "Net" with a couple constants, and functions.

I was imagining the next iteration of this to be to move the Socket module into this module, in some capacity.

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#

### DIFF
--- a/docs/docs/standard-lib/net.md
+++ b/docs/docs/standard-lib/net.md
@@ -1,0 +1,34 @@
+---
+layout: default
+title: IO
+nav_order: 13
+parent: Standard Library
+---
+
+# IO
+{: .no_toc }
+
+## Table of contents
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+## Net
+
+To make use of the Net module an import is required.
+
+```cs
+import Net;
+```
+
+### Net.parseIp4(String) -> Nil
+
+parseIp4 takes a string representation of an IP4 address and returns a Result that unwraps to list containing an entry for each octet.
+
+```cs
+Net.parseIp4("10.0.0.2");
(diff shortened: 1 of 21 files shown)


Jason2605:
Yeah moving Socket into a more generic Net module probably makes more sense. Either that or you make separate ones for each (like Python does: https://docs.python.org/3/library/ipaddress.html / https://docs.python.org/3/library/socket.html)

briandowns:
I'll defer to your judgement on how you want to move forward. Adding Socket to Net had me start thinking about sub modules within a module and how that'd work from an implementation perspective and how it'd look from a language perspective. For the latter, I like any of the following but possibly the "/" most.

```cs
import Net:Socket
```

```cs
import Net/Socket
```

Jason2605:
Yeah submodules will be an interesting one, we can kind of do it at the moment with `from` (although only 1 layer deep).

The only other alternative we could potentially have is:

```
import Net.Socket;
```

Def worth a think on the best syntax for them

Jason2605 (commented):
Couple of little ones

Jason2605:
Thanks for this!

I was thinking as well, I probably should have done it a while ago but I can send you an invite to the org if you'd like? You've done more than enough on the repo now to warrant it! Thank you very much!

briandowns:
> Thanks for this!
> 
> I was thinking as well, I probably should have done it a while ago but I can send you an invite to the org if you'd like? You've done more than enough on the repo now to warrant it! Thank you very much!

Whatever works for you, works for me. :) Thanks!

Issue #694, opened 2023-11-06
[FEATURE] Add kill function to the Process Module


briandowns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

It would be helpful to have a kill function in the Process module to be able to kill running processes with a given signal.

### Describe the solution you'd like

```cs
import Process;

const res = Process.kill(8675309, Process.SIG_KILL);
// Result<Nil>
```

The signal argument would be optional and if one is given, it's override the default of (insert choice here :) )

PR #696, opened 2023-11-07
Process.kill Function

briandowns:
# initial impl of Process.kill

Resolves: #694 

### What's Changed:

Adds a kill function to the Process module as well as docs.

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


Jason2605:
Just checking is this still in draft?

briandowns:
> Just checking is this still in draft?

Ready for review. Not quite sure how to test it though, effectively. I've tested it on Mac, FreeBSD, and Linux manually.

docs/docs/standard-lib/process.md
```diff
@@ -50,3 +50,19 @@ If the external process writes to stdout and you wish to capture the output you
 Process.run(["ls", "-la"]).unwrap();
 print(Process.run(["echo", "test"], true).unwrap()); // 'test'
 ```
+
+### Process.kill(Number, Number -> Optional) -> Result\<Number>
```
Jason2605, line 54: This is saying it unwraps a number but it seems to be returning NIL

src/optionals/process.c
```diff
@@ -268,6 +277,58 @@ static Value runNative(DictuVM* vm, int argCount, Value* args) {
...
+    if (argCount == 2) {
+        if (!IS_NUMBER(args[1])) {
+            runtimeError(vm, "Second argument passed to kill() must be a number");
+            return EMPTY_VAL;
+        }
+
+        signal = AS_NUMBER(args[1]);
+    }
+
+    kill(pid, signal);
```
Jason2605, line 326: Here we could do:

```suggestion
    if (kill(pid, signal) == -1) {
        ERROR_RESULT;
    }
```

Also means we can pass 0 as the signal and check if (child) processes are still running, which is a way we could test this at least on *nix.

Will have a big of a research about windows

briandowns:
I "think" we're all set.

Jason2605 (approved):
Thanks for this!

PR #702, opened 2023-11-20
Add options request

briandowns:
# Add options request

### What's Changed:

Adds the ability to send HTTP OPTIONS requests. I didn't add tests as there's no good way of testing it t the moment. Tested manually.

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


src/optionals/optionals.h
```diff
@@ -12,7 +12,6 @@
 #include "log.h"
 #include "http/http.h"
 #include "path.h"
-#include "c.h"
```
Jason2605, line 15: Not sure if this was intentional but we can probably remove the C module in it's entirety now to be honest. Can be done as a diff PR tho I guess.

Thanks for this!

Jason2605 (approved):

src/optionals/optionals.h
briandowns: It was included twice so I figured I'd remove one. I can get a pr in to remove the c module soon.

PR #724, opened 2024-01-07
Added IsLower and isUpper and wordCount string methods

briandowns:
### What's Changed:

Added IsLower and isUpper string methods. Since these methods are to be ran on single character strings, I returned false if the string length is more than 1 however I can see a case for returning or causing a runtime error. I'll defer to you (@Jason2605) for that final say. We can also create an optional "strings" module that contains extra utility functions and this could go in there. 

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


Jason2605:
I would have done it like you have now to be honest, methods on the string and always return a boolean value (regardless of string length!) I'll get round to reviewing them, sorry for being slow on this mate been a bit hectic in last few days

briandowns:
No rush. Thanks!

docs/docs/strings.md
```diff
@@ -276,4 +276,35 @@ Returns a new string with the original string repeated the given number of times
...
 "ba"+"na".repeat(2); // banana
 "la".repeat(2) + " land"; // lala land
-```
\ No newline at end of file
+```
+
+### string.isUpper() -> Boolean
+
+Returns a boolean indicating that the given string is an upper case letter.
+
```
Jason2605, line 284: I'd probably put a note just explaining empty would return false for `isUpper` and `isLower`

Jason2605:
Actually hold on I think I misunderstood your initial question. You're returning false if the string is greater than 1, not if it's 0. I would expect these methods to still return true if the string length is more than 1 but all characters are lower case, e.g `"dictu".isLower() -> True` (same with `isUpper`) thoughts?

docs/docs/strings.md
briandowns: Updated.

Jason2605 (approved):
Sorry for being slow on this, thank you for doing it!

Issue #726, opened 2024-01-16
[BUG] Adding value to Dictionary leads to hang sometimes


liz3:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Apologies that the code is a bit bigger, but i thought maybe it helps pin down the issue.
This code is a implementation of [Karger's algorithm](https://en.wikipedia.org/wiki/Karger%27s_algorithm) for the [Advent of code coding Puzzle 2023 day 25](https://adventofcode.com/2023/day/25).

Sometimes this section hangs forever either in the first assigment(`n1[rr.first] = 1;` or the second `n2[k] = 1;`):
```
var edgesToMerge = nodes[rr.second].edges;
edgesToMerge.forEach(def (k, v) => {
  print("start forEach call");
  print("1: removing {} from {} in {}".format(rr.second, nodes[k].edges, k));
  nodes[k].edges.remove(rr.second);
  var n1 = nodes[k].edges;
  print("1: adding {} into {} in {}".format(rr.first, n1, k));
  n1[rr.first] = 1;
  var n2 = nodes[rr.first].edges;
  print("2: adding {} into {} in {}".format(k, n2, rr.first));
  n2[k] = 1;
  print("end forEach call");
});
```
But again only sometimes, the other times the program ends successfully.

### Expected Behavior

The program should not lock during the assigment into the Dictionary

### Steps To Reproduce

1. Create input.txt from given file
```
jqt: rhn xhk nvd
rsh: frs pzl lsr
xhk: hfx
cmg: qnr nvd lhk bvb
rhn: xhk bvb hfx
bvb: xhk hfx
pzl: lsr hfx nvd
qnr: nvd
ntq: jqt hfx bvb xhk
nvd: lhk
lsr: lhk
rzs: qnr cmg lsr rsh
frs: qnr lhk lsr
```
2. Create a file from the given source code:
  ```
import Random;
import Queue;
class Node {
  init(var name, var edges) {}
}
class Edge {
  init(var first, var second, var origFirst, var origSecond) {}

}
with("input.txt", "r") {
    var line;
    var lines = [];
    var origNodes = {};
    // When you reach the end of the file, nil is returned
    while((line = file.readLine()) != nil) {
        lines.push(line);
        var parts = line.split(" ");
        parts[0] = parts[0][0:parts[0].find(":")];

        parts.forEach(def (entry) => {
           if(not origNodes.exists(entry)) {
               var n = Node(entry, {});
[...]

PR #728, opened 2024-01-16
fix: correctly return tombstone when theres no free element

liz3:
# fix: correctly return tombstone when theres no free element

The dict findDictEntry function did not correctly handle the case when, a dict had no "free" slot but had tombstones, it would loop forever.

Resolves: https://github.com/dictu-lang/Dictu/issues/726

### What's Changed:
This fixes that by conditionally returning a tombstone when all other indexes where processed.

### Type of Change:

- [x] Bug fix
- [ ] New feature

#

### Housekeeping:

### DIFF
--- a/src/vm/datatypes/dicts/dicts.c
+++ b/src/vm/datatypes/dicts/dicts.c
@@ -31,7 +31,7 @@ static Value lenDict(DictuVM *vm, int argCount, Value *args) {
     }
 
     ObjDict *dict = AS_DICT(args[0]);
-    return NUMBER_VAL(dict->count);
+    return NUMBER_VAL(dict->activeCount);
 }
 
 static Value keysDict(DictuVM *vm, int argCount, Value *args) {
--- a/src/vm/object.c
+++ b/src/vm/object.c
@@ -167,6 +167,7 @@ ObjList *newList(DictuVM *vm) {
 ObjDict *newDict(DictuVM *vm) {
     ObjDict *dict = ALLOCATE_OBJ(vm, ObjDict, OBJ_DICT);
     dict->count = 0;
+    dict->activeCount = 0;
     dict->capacityMask = -1;
     dict->entries = NULL;
     return dict;
--- a/src/vm/object.h
+++ b/src/vm/object.h
@@ -146,6 +146,7 @@ typedef struct {
 struct sObjDict {
     Obj obj;
     int count;
+    int activeCount;
     int capacityMask;
     DictItem *entries;
 };
--- a/src/vm/value.c
+++ b/src/vm/value.c
@@ -77,7 +77,6 @@ static DictItem *findDictEntry(DictItem *entries, int capacityMask,
                                Value key) {
     uint32_t index = hashValue(key) & capacityMask;
     DictItem *tombstone = NULL;
-    uint32_t count = 0;
 
     for (;;) {
         DictItem *entry = &entries[index];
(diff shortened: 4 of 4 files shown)


Jason2605:
We wouldn't want this as there is a potential that the element has wrapped around to the start - will need some more investigation into your issue as to why you're currently getting an infinite loop (more than likely something being freed when it shouldn't have if it's non-deterministic)

liz3:
This condition only applies once all the entries where iterated at least once, so wrapping around should'nt affect it.

Ofc it might be that the dict is shrunken when it should'nt leading to there being no elements "free" as in no key and no value.
So the issue is deterministic but my repro example isn't since the algo is inherently based on randomness(seeing the issue) making the way the dicts resize not deterministic.
But the condition is: some sized dict with no entries where both the key and the value are unset.

Jason2605:
Ah yes sorry you're right the count would have to be the full size of the internal list. Yeah I'd like to do a little more research into this as if this is an issue here (as in finding elements in the dict) it would be an issue with the internal Table type that came from craftinginterpreters which could be a much bigger issue as this type is used throughout the codebase.

Many thanks for the PR and the really detailed issue by the way!

liz3:
@Jason2605 Apologies it took a while i had to do some debugging but heres a deterministic example to correctly and always reproduce the error:
```
var e = {};
e["lkc"] = 1;
e["qvh"] = 1;
e["bck"] = 1;
e["bks"] = 1;
e["bpp"] = 1;
e.remove("bks");
e["gpx"] = 1;
e.remove("bpp");
e["ndj"] = 1;
e.remove("lkc");
e["xmn"] = 1;
e["std"] = 1;
```
Maybe that can help.

Jason2605:
Thank you so much for the simpler reproduction this will help me a lot!!

Jason2605:
Okay yeah, so the issue is the fact that the count is being lowered but tombstones are still technically a "value" and aren't cleared until the dictionary is resized. For dictionaries we probably want a track of `count` and maybe `countActive` so we can expose the length to dictu code

liz3:
> Okay yeah, so the issue is the fact that the count is being lowered but tombstones are still technically a "value" and aren't cleared until the dictionary is resized. For dictionaries we probably want a track of `count` and maybe `countActive` so we can expose the length to dictu code

Can you elaborate maybe on how to solve the issue, since i  see two ways(unrelated of exposing the count as a public api):
* Reuse tombstones more aggresively(This is abstractly seen what the pr does)
* Change the logic to resize the capacity in a way where this condition does not happen.

Jason2605:
The simplest solution in my eyes is to have an extra `count` so `countActive` that keeps tracks of values that are not tombstones and `count` becomes the count of values + tombstones. `count` would still be used the same internally as is and `countActive` is what we'd use for `.len()`

Add `countActive`
https://github.com/dictu-lang/Dictu/blob/develop/src/vm/object.h#L148

`countActive++`
https://github.com/dictu-lang/Dictu/blob/develop/src/vm/value.c#L150

Remove `count--` Add `countActive--`
https://github.com/dictu-lang/Dictu/blob/develop/src/vm/value.c#L162

`countActive`
https://github.com/dictu-lang/Dictu/blob/develop/src/vm/datatypes/dicts/dicts.c#L34

liz3:
Ive changed the pr to reflect the discussed changes, let me know if i missed anything!

Jason2605:
Thanks for this!

PR #730, opened 2024-01-20
feat: add list.sortFunc

liz3:
# feat: list.sortFunc

### What's Changed:
Adds a function to list implemented in dictu which uses quickSort to allow custom sorting which is suuuper useful for a lot of problems. I would have preferred to implement this is C too but there doesn't seam to be a straight forward way to pass callbacks into c. Using a hybrid approach would have been hacky too, but of course this approach does duplicate the sorting logic.

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

### DIFF
--- a/docs/docs/collections/lists.md
+++ b/docs/docs/collections/lists.md
@@ -234,6 +234,20 @@ list1.sort();
 print(list1); // [-1, 1, 2, 3, 4, 5, 10]
 ```
 
+#### list.sortFunc(Func)
+
+Sorts a list with a Custom Callback
+
+```cs
+var list1 = [[1, "Dictu"], [-1, "SomeValue"], [5, "!"], [4, "Awesome"], [2, "is"]];
+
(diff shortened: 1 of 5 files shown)


briandowns:
LGTM

Jason2605 (commented):
Yeah unfortunately due to some optimisations in the VMs main loop it means it's no reentrant so using C wouldn't be possible in it's current state, just one little comment!

liz3:
> Yeah unfortunately due to some optimisations in the VMs main loop it means it's no reentrant so using C wouldn't be possible in it's current state, just one little comment!

Is there something you wan't me to change in the PR, referencing: `just one little comment!`

Jason2605:
Oh what I typed out a comment 😢 it was just a note in the documentation explaining the return value of the callback and how it works essentially

liz3:
@Jason2605 done!

Jason2605:
Legend, thank you for this!

PR #731, opened 2024-01-23
Feat: Buffer api

liz3:
# Buffer API

### What's Changed:
Implements a Buffer API inspired by Node.js's buffer api, including writing strings and all integer types.
Sadly even though i tried to reduce it as far as i could theres a loooot of duplication because of the fact there are 4 integer types, im open to suggestions which can help reduce the duplication.

**Note**: I commited `.clang-format` because thats the formatter i use and it's default setting is set to 2 spaces and so i thought it might be useful, can remove if not needed

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

image


briandowns:
This is super cool. Reminds me that we still need to get that native bytes type implemented. 

As for the PR, I'll refrain from commenting until you're ready but I'll mention there's a few places that still have the original code / names from the copy/paste. I do the exact same thing. :) 

Side note, and happy to take to a different forum like the discussion section but would love to understand your uses of Dictu!

liz3:
I wonder mostly if theres a way to reduce the copy paste for the integer methods which currently take up the majority of the file and only contain very small differences, thinking to have a generalised read/write to reduce it and then just have the typecast to the correct integer type.

Regarding forum you can mail me: liz3@liz3.net or discord: liz3

Jason2605:
Damn yeah this looks really neat! What I'll do with it being WIP is mark it as draft, then just mark it as ready whenever you're good to go!

Jason2605:
Sorry for being slow on this! @briandowns be good to get your thoughts on this one too if you get a spare moment!

tests/buffer/get.du
```diff
@@ -0,0 +1,32 @@
+/**
+* get.du
+*
+* Testing the buffer.get() method
+*
+* .get(Number) returns the given vaue at the index.
```
briandowns, line 6:
```suggestion
* .get(Number) returns the given value at the index.
```

docs/docs/standard-lib/buffer.md
```diff
@@ -0,0 +1,480 @@
...
+
+```cs
+const buffer = Buffer.fromString("Dictu!").unwrap();
+print(buffer);
+// <Buffer>
+```
+
+### Buffer.resize(Number) -> Result\<Number>
+
+Resizes the buffer to the given size, the argument needs to be greater then 0 or the function will return a error.
```
briandowns, line 50:
```suggestion
Resizes the buffer to the given size. The argument needs to be greater than 0 or the function will return a error.
```

docs/docs/standard-lib/buffer.md
```diff
@@ -0,0 +1,480 @@
...
+```cs
+const buffer = Buffer.fromString("Dictu!").unwrap();
+const v = buffer.values();
+print(v); // [68, 105, 99, 116, 117, 33]
+```
+
+
+### Buffer.writeString(Number, String) -> Result\<Nil>
+
+Sets a string into buffer given the starting index, if the string doesn't fit in the buffer a error is returned.
```
briandowns, line 119:
```suggestion
Sets a string into buffer given the starting index. Iif the string doesn't fit in the buffer an error is returned.
```

docs/docs/standard-lib/buffer.md
```diff
@@ -0,0 +1,480 @@
...
+```cs
+const buffer = Buffer.new(6).unwrap();
+buffer.writeString(0, "Dictu!");
+```
+
+### Buffer.readString(Number: start -> Optional, Number: end -> Optional) -> Result\<String>
+
+Returns a String with the optional given start and end parameters, this works very similar to subarray.
```
briandowns, line 127:
```suggestion
Returns a String with the optional given start and end parameters. This works very similar to subarray.
```

docs/docs/standard-lib/buffer.md
```diff
@@ -0,0 +1,480 @@
...
+const buffer = Buffer.new(6).unwrap();
+buffer.writeString(0, "Dictu!");
+const sub = buffer.string(0, buffer.len()-1).unwrap();
+print(sub) // "Dictu"
+```
+
+### Buffer.readUInt64LE(Number) -> Result\<Number>
+
+Returns the u64(unsigned 8 byte integer in little endian) value given the starting index.
+Returns a result with the value or a error.
```
briandowns, line 140:
```suggestion
Returns a result with the value or an error.
```

briandowns: There's a few of these throughout the docs. I won't comment on all of them.

briandowns (commented):
This is such an awesome addition. Only a few small comments. Only other thing, which is a small nit pick, would be to add new line chars to the files that are missing them.

I'm already getting some ideas on what I can do with this!

briandowns (approved):
LGTM! Excited to get this in. :D

docs/docs/standard-lib/buffer.md
briandowns: I think there's an extra "i" in the fix.

liz3: yeye i saw, https://github.com/dictu-lang/Dictu/pull/731/files#diff-e47adca1c96ac76e1106a054a4fcab1a04848982b6b34667870d13c04e2686e4R119

src/optionals/buffer.c
```diff
@@ -0,0 +1,1056 @@
...
+    return newResultSuccess(vm, OBJ_VAL(newBuffer));
+}
+
+ObjAbstract *newBufferObj(DictuVM *vm, double capacity) {
+    ObjAbstract *abstract = newAbstract(vm, freeBuffer, bufferToString);
+    push(vm, OBJ_VAL(abstract));
+
+    Buffer *buffer = ALLOCATE(vm, Buffer, 1);
+    buffer->bigEndian = false;
+    buffer->bytes = calloc(1, capacity);
```
Jason2605, line 936: same here

Jason2605 (commented):
Couple of pretty minor points, thanks so much for the effort on this!

liz3:
> Couple of pretty minor points, thanks so much for the effort on this!

Should be addressed.

liz3:
Why do the tests for *nix likes take 2 hrs? Is this coming from my changes or githubs runners?

briandowns:
My guess is that it's the runners. On Linux, MacOS, and FreeBSD, (locally) the suite runs very quickly.

Jason2605:
Thanks so much for this!

Issue #682, opened 2023-10-20
[FEATURE] Pretty Print


Jason2605:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Extend the new IO module to allow us to pretty print different datatypes

### Describe the solution you'd like

We could actually just piggy back off of the JSON module as this allows us to "stringify" values with a given indentation level

briandowns:
I think it might go better in the Object package.

Jason2605:
Fair point, I’d probably agree

briandowns:
Do we want the user to be able to specify the indentation like they can with the JSON module?

Jason2605:
It would be nice to have yeah!

briandowns:
PR coming shortly then.

PR #732, opened 2024-01-24
add pretty print to object mod

briandowns:
Resolves: #682 

### What's Changed:

# Added pretty print function into the Object module. 

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


docs/docs/standard-lib/object.md
```diff
@@ -61,3 +61,27 @@ This method will return a string of the object's hash value.
 ```cs
 Object.hash("Dictu");
 ```
+
+### Object.prettyPrint(Value, indent -> Optional) -> Result\<String>
+
+This method will return a string representation of the given value.
+
+**NOTE** Strings, dicts, lists, numbers, and nil are valid values for pretty printing at this time.
```
Jason2605, line 69:
```suggestion
**NOTE** Strings, dicts, lists, numbers, booleans and nil are valid values for pretty printing at this time.
```

docs/docs/standard-lib/object.md
```diff
@@ -61,3 +61,27 @@ This method will return a string of the object's hash value.
 ```cs
 Object.hash("Dictu");
 ```
+
+### Object.prettyPrint(Value, indent -> Optional) -> Result\<String>
```
Jason2605, line 65:
```suggestion
### Object.prettyPrint(Value, Number: indent -> Optional) -> Result\<String>
```

docs/docs/standard-lib/object.md
```diff
@@ -61,3 +61,27 @@ This method will return a string of the object's hash value.
...
+
+```cs
+Object.prettyPrint([1, 2, 3]).unwrap();
+
+// Output
+'[1, 2, 3]'
+```
+
+```cs
+Object.prettyPrint({"a": 1}, 4).unwrap()
```
Jason2605, line 79:
```suggestion
Object.prettyPrint({"a": 1}, 4).unwrap();
```

src/optionals/object/object.c
```diff
@@ -59,7 +59,21 @@ static Value objectHash(DictuVM *vm, int argCount, Value *args) {
...
+
+    return OBJ_VAL(stringify(vm, argCount, args));
+}
+
 Value createObjectModule(DictuVM *vm) {
+    ObjString *name = copyString(vm, "Object", 6);
+    push(vm, OBJ_VAL(name));
+    ObjModule *module = newModule(vm, name);
+    push(vm, OBJ_VAL(module));
+
```
Jason2605, line 76: We don't need any of this, we have access to the module through the closure

src/optionals/object/object.c
```diff
@@ -75,6 +89,8 @@ Value createObjectModule(DictuVM *vm) {
     defineNative(vm, &closure->function->module->values, "getClassRef", objectGetClassRef);
     defineNative(vm, &closure->function->module->values, "hash", objectHash);
 
+    defineNative(vm, &module->values, "prettyPrint", objectPrettyPrint);
```
Jason2605, line 92:
```suggestion
    defineNative(vm, &closure->function->module->values, "prettyPrint", objectPrettyPrint);
```

Jason2605 (commented):
Couple of comments!

Jason2605:
A question for you, would you expect a method called `prettyPrint` to actually output to stdout rather than return a string? I think a method that returns the pretty output as a string would be useful but potentially under a different name, what do you think?

briandowns:
Yeah, that's a good point. I'll update this to print to stdout. What would you want the method name to be that would returned the "pretty" string?

Jason2605:
Actually we probably don't need the non-print version as you can just use `JSON.stringify` - at least for the minute until we extend it so we can pretty print more objects that aren't JSON serialisable

briandowns:
Good point.

Jason2605:
Thanks so much for this!

PR #733, opened 2024-01-24
remove unneeded file

briandowns:
### What's Changed:

Removes an unneeded file from the root of the repository.

### Type of Change:

- [x] Bug fix
- [ ] New feature

#

### Housekeeping:

#


Jason2605:
We really need to look at that test, it seems to be failing on each run

briandowns:
Really strange how it's fine in one test and then consistently blows up in others, after having been fine for awhile.

Jason2605:
Hmmm looks as though this extended sleep is running forever on mac. Potentially work marking it as skipped for now and we can try look into it

briandowns:
yeah, that was a initial test and then my flight's WiFi went out so I had to wait a bit. I'll skip for now while I keep digging.

Jason2605:
Ah no worries at all mate! Thanks for the cleanup

PR #738, opened 2024-02-10
FFI API

liz3:
# FFI Api for dictu.

### What's Changed:
Adds a FFI api which allows a C abi to easy extend dictu with native functions.

* It works by providing a header(`dictu-include.h`) which contains all macros which might be useful for external users, although of course theres the chance i missed something. 
* The header replaces the compiler entry with a void pointer to avoid having to copy all its structs
* The header contains function pointers for all the required functions from the vm, which are loaded at runtime automatically(`dictu_internal_ffi_init`), this avoids the user having to link against dictu or duplicate function definitions.
* It also contains a mechanism to ensure that function pointers can safely be loaded with a versioning system.
* The pr also adds support for two things: 
   1. Invoking functions on a abstract without self as 0 arg, this is because the remote ffi function should only receive the actual function arguments, the 0 argument is the FFIInstance
   2. Adding properties to abstract which was a quick edit in the vm but might be useful for external modules.

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:


liz3:
im not sure why the docs fail but maybe its the empty links, but all other tests pass.
thats good

liz3:
> Apologies been busy, I def will get round to this, thank you so much!
> 
> Be good to get your eyes too if you have any spare time @briandowns 🙂

All good no stress

.gitignore
```diff
@@ -30,4 +30,5 @@ test1.du
 .vscode
 /.env
 /vcpkg_installed
-/Release
\ No newline at end of file
+/Release
+examples/ffi-example/build
```
briandowns, line 34: Can you add new line characters to any file that don't have them?

examples/ffi-example/ffi.du
```diff
@@ -0,0 +1,14 @@
+import FFI;
+import System;
+
+var mod = nil;
+if(System.platform == "windows") {
+ mod = FFI.load("./build/Release/ffi-test{}".format(FFI.suffix));
```
briandowns, line 6: Looks like some of the spacing is 1, 2, or 4. Can you make them all 4 spaces?

src/optionals/ffi/ffi.c
```diff
@@ -0,0 +1,181 @@
...
+    char *path;
+} FFIInstance;
+
+typedef struct _vm_external vm_external;
+
+typedef Value function_definition_t(DictuVM *vm, int argCount, Value *args);
+typedef int init_func_definition_t(void **function_ptrs, DictuVM *vm,
+                                   Table *table, int vm_ffi_version);
+
+// #define AS_FFI_INSTANCE(v) ((FFIInstance *)AS_ABSTRACT(v)->data)
```
briandowns, line 23: Can we remove this line?

briandowns:
I love this. Very very cool. Few things:

* Can we add the shared object libraries to the .gitignore?
* How are would it be to use Dictu as an embedded language now with the shared objects?
* Even thought it's FFI, I was curious aabout folks thoughts on naming the module "external" or something similar.

I can't wait to get started using this!

liz3:
> I love this. Very very cool. Few things:
> 
>     * Can we add the shared object libraries to the .gitignore?
> 
>     * How are would it be to use Dictu as an embedded language now with the shared objects?
> 
>     * Even thought it's FFI, I was curious aabout folks thoughts on naming the module "external" or something similar.
> 
> 
> I can't wait to get started using this!

> Can we add the shared object libraries to the .gitignore?

Uh if you can help me build the test libs with the runner sure.

> How are would it be to use Dictu as an embedded language now with the shared objects?

Im not entirely sure what you mean by embedded here, as in including shared objects into a binary? Im a tad confused.

> Even thought it's FFI, I was curious aabout folks thoughts on naming the module "external" or something similar.

We can do that idm about the name at all.

examples/ffi-example/src/dictu-include.h
```diff
@@ -0,0 +1,643 @@
+#ifndef dictumod_include_h
```
Jason2605, line 1: It's probably worth adding this to the include directory within src, that way people may find it easier rather than it being in the examples directory

Jason2605:
This is really really awesome!!! Thank you so much for working on this!! I'm gonna merge this in but left a little comment about moving / copying one file that any of us can pick up at some point (not massively urgent)

Jason2605 (approved):

Issue #317, opened 2020-11-07
Add unicode support


Jason2605:
# Add unicode support
## Summary
Currently within Dictu all string based activities are done on the number of bytes a string contains, rather than the number of characters, obviously this means some non-ascii characters will be dealt with wrong. We should change this to instead include support for unicode characters within the default strings.

## Example
```
Dictu Version: 0.11.0
>>> "Ā".len();
2
```

Jason2605:
A potentially promising library: https://github.com/sheredom/utf8.h

liz3:
<img width="798" alt="image" src="https://github.com/user-attachments/assets/b0af1e67-41d8-47ec-ad25-890eda103514">
👀

Jason2605:
[...]

PR #749, opened 2024-09-18
utf-8 support in strings

liz3:
# Add Unicode support to strings

Resolves: https://github.com/dictu-lang/Dictu/issues/317

### What's Changed:
This updates the implementation of the string API to support unicode using the [
utf8.h](https://github.com/sheredom/utf8.h) library.

Theres still a few incomplete things, some fuctions do not need unicode versions(strip functions). But on the other side there are things where the library does not provide full functionality which is needed for `isUpper`/`isLower` for example which do not wrong, forcing the usage of the c std lib apis.

Should `String.len()` return byte length or character length, what if the string has invalid utf8? calculating the length upfront for large strings, always might be expensive.

And not enough tests are added.

### Type of Change:

- [ ] Bug fix
- [X] New feature
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#

### Housekeeping:

#

### DIFF
--- a/src/vm/datatypes/strings.c
+++ b/src/vm/datatypes/strings.c
@@ -702,17 +702,14 @@ static Value isUpperString(DictuVM *vm, int argCount, Value *args) {
         runtimeError(vm, "isUpper() takes no arguments (%d given)", argCount);
         return EMPTY_VAL;
     }
-    {
-        ObjString *string = AS_STRING(args[0]);
-        if(string->character_len == -1) {
+
+    ObjString* strObj = AS_STRING(args[0]);
+    if(strObj->character_len == -1) {
            runtimeError(vm, "String contains invalid UTF-8", argCount);
            return EMPTY_VAL;
-       }
     }
-
-
-    char *string = AS_CSTRING(args[0]);
-    int len = utf8len(string);
+    char *string = strObj->chars;
+    int len = strObj->character_len;
 
     if (len == 0) {
         return BOOL_VAL(false);
@@ -734,16 +731,14 @@ static Value isLowerString(DictuVM *vm, int argCount, Value *args) {
         runtimeError(vm, "isLower() takes no arguments (%d given)", argCount);
         return EMPTY_VAL;
     }
-    {
-        ObjString *string = AS_STRING(args[0]);
-        if(string->character_len == -1) {
-           runtimeError(vm, "String contains invalid UTF-8", argCount);
-           return EMPTY_VAL;
-       }
+    ObjString *strObj = AS_STRING(args[0]);
+    if(strObj->character_len == -1) {
+       runtimeError(vm, "String contains invalid UTF-8", argCount);
+       return EMPTY_VAL;
     }
 
-    char *string = AS_CSTRING(args[0]);
-    int len = utf8len(string);
+    char *string = strObj->chars;
+    int len = strObj->character_len;
 
     if (len == 0) {
         return BOOL_VAL(false);
@@ -793,7 +788,7 @@ static Value collapseSpacesString(DictuVM *vm, int argCount, Value *args) {
     if (targetLen != string->length) {
         dest = SHRINK_ARRAY(vm, dest, char, string->length + 1, targetLen + 1);
     }
-    dest[targetLen] = 0x00;
+    dest[targetLen] = '\0';
 
     return OBJ_VAL(takeString(vm, dest, targetLen));
 }
@@ -810,7 +805,7 @@ static Value wrapString(DictuVM *vm, int argCount, Value *args) {
        return EMPTY_VAL;
    }
     char *temp = ALLOCATE(vm, char, string->length + 1);
-    temp[string->length] = 0x00;
+    temp[string->length] = '\0';
     memcpy(temp, string->chars, string->length);
     int len = AS_NUMBER(args[1]);
 
--- a/src/vm/datatypes/strings.c
+++ b/src/vm/datatypes/strings.c
@@ -803,7 +803,7 @@ static Value wrapString(DictuVM *vm, int argCount, Value *args) {
     if(string->character_len == -1) {
        runtimeError(vm, "String contains invalid UTF-8", argCount);
        return EMPTY_VAL;
-   }
+    }
     char *temp = ALLOCATE(vm, char, string->length + 1);
     temp[string->length] = '\0';
     memcpy(temp, string->chars, string->length);
--- a/src/vm/datatypes/strings.c
+++ b/src/vm/datatypes/strings.c
@@ -153,11 +153,16 @@ static Value splitString(DictuVM *vm, int argCount, Value *args) {
 
     ObjString *string = AS_STRING(args[0]);
     if(string->character_len == -1) {
-     runtimeError(vm, "String contains invalid UTF-8", argCount);
+     runtimeError(vm, "String contains invalid UTF-8");
      return EMPTY_VAL;
      }
 
-    char *delimiter = AS_CSTRING(args[1]);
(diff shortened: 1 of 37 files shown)


Jason2605:
This looks awesome and something definitely needed, thank you for this!! Sorry I've been so slow on this, been a bit hectic for me as of late!

> Should String.len() return byte length or character length

I would expect this to return the character length not the byte length. I like the `byteLen` addition! I'm more than happy with adding some extra test cases!

liz3:
> This looks awesome and something definitely needed, thank you for this!! Sorry I've been so slow on this, been a bit hectic for me as of late!
> 
> > Should String.len() return byte length or character length
> 
> I would expect this to return the character length not the byte length. I like the `byteLen` addition! I'm more than happy with adding some extra test cases!
> 
> We will need to make sure this also works with the stdlib modules too, quite a big change!!

Hey, ive continued on this. I think the changes are needed, but i want to make clear it introduces a overhead into indexing/slicing and allocation of strings since simple indexing does not work anymore and is iterative now, the other approach is to store the strings as unicode in memory, this would imbue a massive increase in memory usage but retain indexing capabilities, im not entirely sure what the better tradeoff is here.

Ive went with approaching that most string functions will throw errors when provided with invalid utf-8, i do not know how the library would behave when given invalid utf-8. Exceptions are indexing and slicing which will fallback to the byte handling.
It might also be worth adding a function which lets the user know if the string is valid utf-8 maybe?

liz3:
I am not entirely sure why the windows tests fail, there does not seam to be a clear test which **is** failing
Sadly at the moment i don't have access to a Window system.

liz3:
@Jason2605 I believe this is good for review / merge now. I have fixed the issue with the windows builds, which turned out to be a error on my side but it took a good while to find. I see that theres a macos failure but i have run the tests on my mb and they all pass.

Jason2605:
Nice one thank you! I don't have an ARM mac myself to exactly test unfortunately but will have a see if I can track someone down that does.

@briandowns be good to get your thoughts on this PR too!

liz3:
> Nice one thank you! I don't have an ARM mac myself to exactly test unfortunately but will have a see if I can track someone down that does.
> 
> @briandowns be good to get your thoughts on this PR too!

i have a arm mac, infact i only have apple sillicon machines

liz3:
The way runners fail also make it hard to see the error. The log does not really give you a hint or so to start debugging, but i ran all the tests on m2 air last night

Jason2605:
> The way runners fail also make it hard to see the error. The log does not really give you a hint or so to start debugging, but i ran all the tests on m2 air last night

Oh interesting I see. Yeah the logs are no help at all really 😂 Usually when it's just a dump on code 1 it's a segfault somewhere, so will need to see if we can track it

liz3:
> > The way runners fail also make it hard to see the error. The log does not really give you a hint or so to start debugging, but i ran all the tests on m2 air last night
> 
> Oh interesting I see. Yeah the logs are no help at all really 😂 Usually when it's just a dump on code 1 it's a segfault somewhere, so will need to see if we can track it

Aight maybe then a adresssanitizer can help to find it if its a heap corruption, Il try that

liz3:
@Jason2605 was able to repro with a debug build and fixed it: https://github.com/dictu-lang/Dictu/pull/749/commits/f8d481843ba1c1529f572b8dc378f58dbd17b053

liz3:
image
okay that's satisfying to be entirely honest!

Jason2605:
Incredible stuff, this is so so so nice!!!!!

src/vm/datatypes/strings.c
```diff
@@ -147,43 +171,44 @@ static Value splitString(DictuVM *vm, int argCount, Value *args) {
...
+        for (; tokenIndex < string->character_len && count < maxSplit; tokenIndex++) {
             count++;
-            *(tmp) = string->chars[tokenIndex];
-            *(tmp + 1) = '\0';
-            Value str = OBJ_VAL(copyString(vm, tmp, 1));
+          
+            char* next_ptr = utf8codepoint(ptr, &cp);
+
+            // *(tmp) = string->chars[tokenIndex];
+            // *(tmp + 1) = '\0';
```
Jason2605, line 191:
```suggestion
```

src/vm/datatypes/strings.c
```diff
@@ -239,7 +270,13 @@ static Value findString(DictuVM *vm, int argCount, Value *args) {
         runtimeError(vm, "find() takes either 1 or 2 arguments (%d given)", argCount);
         return EMPTY_VAL;
     }
-
+    {
```
Jason2605, line 273: Do we need the extra scope here? It seems to happen throughout quite a few times

src/vm/datatypes/strings.c
```diff
@@ -209,16 +234,22 @@ static Value splitString(DictuVM *vm, int argCount, Value *args) {
...
 
 static Value containsString(DictuVM *vm, int argCount, Value *args) {
+
     if (argCount != 1) {
         runtimeError(vm, "contains() takes 1 argument (%d given)", argCount);
         return EMPTY_VAL;
     }
-
+    {
+        ObjString *string = AS_STRING(args[0]);
```
Jason2605, line 247: If we're checking utf8 legality would we also need to check the needle (args[1])?

src/vm/datatypes/strings.c
```diff
@@ -280,42 +319,34 @@ static Value findLastString(DictuVM *vm, int argCount, Value *args) {
...
+        if(utf8str(ptr, ss->chars) == ptr)
+            return NUMBER_VAL(index);
+        index--;
+        if(index < 0)
+            break;
+        ptr = utf8rcodepoint(ptr, &cp);
     }
 
-    return NUMBER_VAL(found ? p-str : -1);
+    return NUMBER_VAL( -1);
```
Jason2605, line 349:
```suggestion
    return NUMBER_VAL(-1);
```

src/vm/datatypes/strings.c
```diff
@@ -701,31 +800,51 @@ static Value wrapString(DictuVM *vm, int argCount, Value *args) {
     }
 
     ObjString *string = AS_STRING(args[0]);
+    if(string->character_len == -1) {
+       runtimeError(vm, "String contains invalid UTF-8", argCount);
+       return EMPTY_VAL;
+   }
```
Jason2605, line 806:
```suggestion
    }
```

src/vm/datatypes/strings.c
```diff
@@ -701,31 +800,51 @@ static Value wrapString(DictuVM *vm, int argCount, Value *args) {
...
-    for (int cur = 0; string->chars[cur] != '\0'; cur++, count++) {
-        temp[cur] = string->chars[cur];
-
-        if (isspace(temp[cur])) {
-            last = cur;
+    char*ptr = temp;
+    for (int cur = 0; cur < string->character_len; cur++, count++) {
+        utf8_int32_t cp;
+        char* n = utf8codepoint(ptr, &cp);
+        // temp[cur] = string->chars[cur];
```
Jason2605, line 818:
```suggestion
```

Jason2605 (commented):
Few comments, most are nitpick style ones. Again thank you so much for this, I appreciate the effort that has gone into this :)

src/vm/datatypes/strings.c
liz3: I have refactored them to not do that anymore, it was for name shadowing but the fix was pretty easy

liz3:
@Jason2605 i ran clang-format over the file in order to fix indentation

Jason2605:
Noice yeah I should probably add something to the GH actions for that!

Very much appreciated, once again thank you for this!

PR #753, opened 2024-10-30
fix: when executing a child process, do not waitpid before reading

liz3:
Trying to waitpid before reading the piped content will lead to a blocked pipe and get the child process "stuck". So it should be read BEFORE waitpid is called. Further there was a bug within the reallocation logic leading to heap corruption because the comparison size did not include the just read chunk, leading to heap corruption, which itself led to reallocate failing.

### Type of Change:

- [x] Bug fix
- [ ] New feature

#

### Housekeeping:


Jason2605:
Nice, thank you!

PR #755, opened 2024-11-13
feat: call dictu from native code

liz3:
# feat: call dictu from native code

This uses a stack loaded piece of code to interrupt the current vm execution in order to invoke a function called by native code.

It creates a new callframe for that and then makes sure the vm loop breaks after that frame was reached again, this way the result is returned to the caller. It then resets the vm state and continues normally.

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#

```
const entry = "test";

def testFunc (a,b=30) {
  print("Hello world {}".format(entry));
  if(a == 30)
    return nil;
   return a+b;
}

print(entry.testThing(
testFunc, 30
));

```


liz3:
@Jason2605 would be super thrilled to get some feedback here

src/vm/vm.c
```diff
@@ -2457,3 +2469,21 @@ DictuInterpretResult dictuInterpret(DictuVM *vm, char *moduleName, char *source)
 
     return result;
 }
+Value callFunction(DictuVM* vm, Value function, int argCount, Value* args) {
+    if(!IS_FUNCTION(function) && !IS_CLOSURE(function))
+        return NIL_VAL;
+    int currentFrameCount = vm->frameCount;
+    Value* currentStack = vm->stackTop;
+    CallFrame *frame = &vm->frames[vm->frameCount++];
```
Jason2605, line 2477: Here we need to be careful we actually have enough frames, we may need to reallocate if not

src/vm/vm.c
```diff
@@ -2457,3 +2469,21 @@ DictuInterpretResult dictuInterpret(DictuVM *vm, char *moduleName, char *source)
 
     return result;
 }
+Value callFunction(DictuVM* vm, Value function, int argCount, Value* args) {
+    if(!IS_FUNCTION(function) && !IS_CLOSURE(function))
```
Jason2605, line 2473: Here it would be nice to return an EMPTY val and throw a runtime error (happy you're just testing so may not have put that in yet etc)

Jason2605: Also be nice to handle natives here too, that should be a lot easier as we don't need to handle call frames

src/vm/vm.c
```diff
@@ -2457,3 +2469,21 @@ DictuInterpretResult dictuInterpret(DictuVM *vm, char *moduleName, char *source)
...
+    int currentFrameCount = vm->frameCount;
+    Value* currentStack = vm->stackTop;
+    CallFrame *frame = &vm->frames[vm->frameCount++];
+    uint8_t code[4] = {OP_CALL, argCount, 0, OP_RETURN};
+    frame->ip = code;
+    push(vm, function);
+    for(int i = argCount -1; i >= 0; i--) {
+        push(vm, args[i]);
+    }
+    runWithBreakFrame(vm, currentFrameCount+1);
```
Jason2605, line 2484: Here we would need to handle a runtime error as at the minute (without testing) I'm assuming this will probably dump

Jason2605 (commented):
This seems really promising so far! A couple of comments to think about but this would be huge 😄

liz3:
@Jason2605 I've addressed the review points.
The next step will be adding tests but i am yet to decide how to do that.

I am very thankful you do not seam against this approach because i was very uncertain whether this would have other side effects but it seams that it does give a good way to callback into dictu code from natives, the overhead is manageable i hope.

liz3:
added tests!

Jason2605:
I think we're almost there! I think the runtime stack trace does segfault out though if a runtime error is actually raised, so we will need to have a look into that

liz3:
@Jason2605 ive fixed the issue, i had to read the error logic to make sense of the issue but after that it appeared logical.
I intentionally set the closure to NULL and then have a special case within the error stack trace, should there be more "types of callframes" in the future we might need to change the approach but for now this should work

Jason2605:
I actually wonder if it's worth silencing it for the synthetic callframe. I'm not sure it gives the end user a whole lot of benefit and may cause confusion, what do you think?

Screenshot 2024-11-15 at 00 46 02

Screenshot 2024-11-15 at 00 45 28

liz3:
Silencing should be fine too

liz3:
http test service returning 500 :melting_face:

Jason2605:
Aha yeah the tests rely on an external service (not great). I’ve re-ran them now

Jason2605:
It’s 1am for me right now so I’ll give this a proper look over tomorrow! Thank you again for this :)

tests/ffi/ffi.du
```diff
@@ -38,18 +38,51 @@ import FFI;
 import Path;
 
 class TestFFIModule < UnitTest {
+    var mod = nil;
+    doSetup() {
```
Jason2605, line 42:
```suggestion
    setUp() {
```

tests/ffi/ffi.du
```diff
@@ -38,18 +38,51 @@ import FFI;
...
+        const path = Path.join(parts);
+        this.mod = FFI.load(path);
+    }
     testFFIModule() {
-        const path = Path.join(Path.dirname(__file__), "libs", "test-lib{}{}".format(
-            System.platform == "darwin" ? System.arch == "x86_64" ? "_64" : "_arm" : "",
-            FFI.suffix
-        ));
-        const mod = FFI.load(path);
+        this.doSetup();
```
Jason2605, line 57:
```suggestion

```

tests/ffi/ffi.du
```diff
@@ -38,18 +38,51 @@ import FFI;
...
+        const mod = this.mod;
         this.assertEquals(mod.test, "Dictu!");
         for(var i = 0; i < 40; i+=2) {
             this.assertEquals(mod.dictuFFITestAdd(i, i), i+i);
         }
         this.assertEquals(mod.dictuFFITestStr(), "Hello From Dictu FFI module!");
     }
+    testFFIModuleCallback() {
+        this.doSetup();
```
Jason2605, line 66:
```suggestion

```

Jason2605 (commented):
We have the `setUp` method we can use here that is implicitly called before each test run!

https://dictu-lang.com/docs/standard-lib/unittest/#setup

liz3:
> We have the `setUp` method we can use here that is implicitly called before each test run!
> 
> https://dictu-lang.com/docs/standard-lib/unittest/#setup

done

Jason2605:
God damn HTTP tests 😠

Jason2605:
It's getting merged. Good find with this! Thank you for the PR 😃

Issue #670, opened 2023-09-29
[FEATURE] Type Hints (Discussion)


Jason2605:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Building out a fully blown type system is probably a little above my pay grade, however, what we can potentially do is add type hints into dictu. These would function exactly the same as Python's type hints (which are essentially glorified comments) but will allow us to use external tools to check for things, or even just provide a little more clarity to the developer about a type even if there isn't actually any checking taking place

### Describe the solution you'd like

Something along the lines of:

```
const myString: string = "hello!!";
const myString -> string = "hello!!";

def lengthGreatherThanFive(myString: string): boolean {
    return myString.len() > 5;
}

def lengthGreatherThanFive(myString -> string) -> boolean {
    return myString.len() > 5;
}

// Maybe a mix?
def lengthGreatherThanFive(myString: string) -> boolean {
    return myString.len() > 5;
}
```

briandowns:
Definitely in favor of the `:` option.

RevengerWizard:
Python's way of handling type hints is especially tricky and weird, given that it treats everything after the `:` or `->` as a normal language expression that doesn't produce any bytecode, essentially a comment. 

It's also interesting how Typescript, but usually Javascript, handle type hints inside separate `.d.ts` files which basically only contain the relevant hints to functions, classes and attributes.

PR #757, opened 2024-11-15
Type Hints

Jason2605:
# Type Hints

Resolves: #670 

### What's Changed:

- Allow the compiler to recognise (and ignore) type hints. This will allow third party type checkers to be created

#

### Type of Change:

- [ ] Bug fix
- [x] New feature

#

### Housekeeping:

#


Jason2605:
Dunno what your guys thoughts are on this @liz3 @briandowns ?

Would allow us to create a static analyser outside of the Dictu codebase (potentially written in Dictu :O) similar to MyPy and others

liz3:
I don't mind something like that as long as its optional. Good static analysers are able to do with without hints by inferring based on default values etc.
Amongs my peers i have the unpopular opinion of **not** liking typescript for example. Imo it does not add actual value since most bugs are logic errors which a static analyser cannot catch. 
That being said i do like them once the vm(runtime) can use them to provide a performance advantage through making assumptions it cannot do otherwise.

The TL'DR for me is: I think in the current form(optional) they are okay but nothing i would put too much value on.

briandowns:
I'm with @liz3 . I don't think I'd use them much unless I'm writing a huge code base and it'd be a pain to track back through a ton of files to figure out what's what. Super cool feature though from a parser / language perspective though.

PR #758, opened 2024-11-16
multiple fixes of small things and QOL

liz3:
# multiple fixes of small things and QOL

### What's Changed:
* Fixes `mkdirAll` which would fail if one of the directories would already exist.
* adds `eprint` and` eprintln` to the IO module
* add a CMake option to **not** relocate the main executable(i use this to have two builds debug/not debug at the same time`
* Adds a git hash and debug info to the version
* fixes some stuff in the docs

#

### Type of Change:

- [x] Bug fix
- [x] New feature

#

### Housekeeping:

#


liz3:
@briandowns @Jason2605

briandowns (approved):
LGTM

tests/path/isDir.du
```diff
@@ -3,7 +3,7 @@
  *
  * Testing Path.isDir()
  *
- * Returns true if the given string is a path to a directory, else false. (Linux only) 
+ * Returns true if the given string is a path to a directory, else false.`
```
liz3, line 6:
```suggestion
 * Returns true if the given string is a path to a directory, else false.

```

Jason2605:
Nice, thanks for this!

PR #766, opened 2026-02-15
Fibers

### LINKED ISSUE #87: STACK_MAX is a constant
`STACK_MAX` is a constant and there is no checking in `push()` to see if we have hit the stack limit. This is for good reason, as `push()` happens so frequently, a branch in here would hit performance significantly, however this does mean the VM can run into undefined behaviour if the stack count out grows the hard constant. This should instead be a value based on the "worst case" amount of values worked out at compile time.

### LINKED ISSUE #487: [FEATURE] lightweight user-level concurrency
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Many languages now provide a lightweight user-level concurrency mechanisms, from fibers in Wren to goroutines in Go or async callbacks in JavaScript (the latter being the least usable of the three). Adding this to Dictu would enable instructors to teach concurrency.

### Describe the solution you'd like

Co-operative concurrency (e.g., fibers in Wren) is straightforward to implement and will be familiar to most instructors.

### Describe alternatives you've considered

1. JavaScript-style callbacks and/or promises have proven very hard to teach.
2. Linda-style tuple spaces are easy to teach and understand, but the pattern matching would require larger changes to Dictu.
3. Software Transactional Memory (STM) is easier to understand than locks or monitors (https://neverworkintheory.org/2011/08/24/is-transactional-programming-actually-easier.html) but again, would require larger changes to the VM and language than fibers.

### Additional context

I'm looking for a small teaching language (think Pascal in the early 80s, or Python in the early 2000s); Dictu seems like a very promising candidate. I'm extracting requirements from https://third-bit.com/sdxjs/ and https://third-bit.com/sdxpy/.

Jason2605:
# Fibers

Resolves: #487
Resolves: #87 

### What's Changed:

Implements Fibers (https://github.com/wren-lang/wren used as base) into the language
Also fixes an issue where the VM couldn't grow

#

### Type of Change:

- [x] Bug fix
- [x] New feature

#

### Housekeeping:

#

@briandowns not sure if you're still watching, but I think this is quite exciting!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ build_x64/
 sessionmanager-bundle.zip
 
 sessionmanager-bundle/
+
+build_debug/
--- a/docs/docs/standard-lib/fiber.md
+++ b/docs/docs/standard-lib/fiber.md
@@ -1,0 +1,167 @@
+---
+layout: default
+title: Fiber
+nav_order: 8
+parent: Standard Library
+---
+
+# Fiber
+{: .no_toc }
+
+## Table of contents
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+## Fiber
+
+Fibers are lightweight cooperative coroutines. Each fiber has its own call stack, and execution can be suspended and resumed at will. They are useful for generators, pipelines, and cooperative multitasking.
+
+To make use of the Fiber module an import is required.
+
+```js
+import Fiber;
+```
+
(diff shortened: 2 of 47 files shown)


briandowns:
This is incredible work.

Jason2605:
Thank you! Will merge it in and probably start work on trying to add some non-blocking features in to truly make use

